PR #115, opened 2014-11-05
start and end date support for every (and more)

elmurci:
Hi!
First of all, thanks for this awesome module. It's great.
Due to my project needs, I have been forced to fork and add a couple of changes:

- every: start and end date support for job executions added, jobs created with every will only run in the date-time interval (if) given. This is helpful as it saves calls to external api's.
- Configuration file conf.js created
- single job types: two jobs with same name but different data will generate two separate records.

Thanks

PR #132, opened 2015-01-02
Formatting & code cleanup

ilanbiala:
Cleaning up some code.

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -8,9 +8,7 @@ var Agenda = module.exports = function(config) {
   if (!(this instanceof Agenda)) {
     return new Agenda(config);
   }
-  if (!config) {
-    config = {};
-  }
+  config = config ? config : {};
   this._name = config.name;
   this._processEvery = humanInterval(config.processEvery) || humanInterval('5 seconds');
   this._defaultConcurrency = config.defaultConcurrency || 5;
@@ -88,7 +86,7 @@ Agenda.prototype.create = function(name, data) {
 Agenda.prototype.jobs = function() {
   var args = Array.prototype.slice.call(arguments);
 
-  if (args.length && typeof args[args.length - 1] === 'function') {
+  if (typeof args[args.length - 1] === 'function') {
     args.push(findJobsResultWrapper(this, args.pop()));
   }
 
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -5,8 +5,12 @@ var Job = require('./job.js'),
   mongo = require('mongoskin');
 
 var Agenda = module.exports = function(config) {
-  if(!(this instanceof Agenda)) return new Agenda(config);
-  if(!config) config = {};
+  if (!(this instanceof Agenda)) {
+    return new Agenda(config);
+  }
+  if (!config) {
+    config = {};
+  }
   this._name = config.name;
   this._processEvery = humanInterval(config.processEvery) || humanInterval('5 seconds');
   this._defaultConcurrency = config.defaultConcurrency || 5;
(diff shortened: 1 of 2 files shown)


ilanbiala:
@rschmukler you should try to use a formatter in whatever editor you use. Also, why so many function declarations inside functions?

rschmukler:
@ilanbiala I use a formatter, however I suspect my JSLinting settings are less strict than the ones you use.  If you'd like, you [can see my jslinting preferences (https://github.com/rschmukler/dotfiles/blob/master/jshintrc).

As for the inline function declarations - it depends. Sometimes I use them to clarify what's happening in the code (ie. a function that calls 4 25 line functions that have a name that describes what they are doing, is often easier to understand than one 100 line function). Other times (specifically in the jobQue related methods) things are used as callbacks multiple times, and as such are needed to be isolated.

Regarding your changes, I like most of them, as I agree, they do help the readability / maintainability of the code. I'm going to comment on a few of them that I don't like, and then if you change them, I will merge. Also, you should squish your commits into one.

Hopefully this doesn't screw up @elmurci's separating of his PRs in [#115]...

lib/agenda.js
```diff
@@ -83,7 +88,7 @@ Agenda.prototype.create = function(name, data) {
 Agenda.prototype.jobs = function() {
   var args = Array.prototype.slice.call(arguments);
 
-  if(typeof args[args.length - 1] == 'function') {
+  if (args.length && typeof args[args.length - 1] === 'function') {
```
rschmukler: unnecessary since `typeof undefined != 'function'`

lib/agenda.js
```diff
@@ -112,15 +117,14 @@ Agenda.prototype.define = function(name, options, processor) {
 Agenda.prototype.every = function(interval, names, data) {
   var self = this;
 
-  if (typeof names === 'string') {
+  if (typeof names === 'string' || names instanceof String) {
```
rschmukler: when would the first one evaluate false, but the second one true? Unless there is a good reason, I'd get rid of this second case.

lib/agenda.js
```diff
@@ -138,7 +142,7 @@ Agenda.prototype.every = function(interval, names, data) {
 Agenda.prototype.schedule = function(when, names, data) {
   var self = this;
 
-  if (typeof names === 'string') {
+  if (typeof names === 'string' || names instanceof String) {
```
rschmukler: See note above

lib/job.js
```diff
@@ -193,9 +196,15 @@ Job.prototype.run = function(cb) {
 };
 
 Job.prototype.isRunning = function() {
-  if (!this.attrs.lastRunAt) return false;
-  if (!this.attrs.lastFinishedAt) return true;
-  if (this.attrs.lockedAt && this.attrs.lastRunAt.getTime() > this.attrs.lastFinishedAt.getTime()) return true;
+  if (!this.attrs.lastRunAt) {
```
rschmukler: Keep the first two as one liners since they're so easy to read. The 3rd one can stay as a multi-liner as you did it.

lib/job.js
```diff
@@ -227,7 +236,7 @@ function parsePriority(priority) {
     high: 10,
     highest: 20
   };
-  if(typeof priority == 'number')
+  if (typeof priority === 'number' || priority instanceof Number)
```
rschmukler: See note above on typeof vs instanceof

lib/agenda.js
ilanbiala: `new String()`, just trying to cover all the bases.

lib/agenda.js
ilanbiala: Right because `a[-1]`is just undefined.

ilanbiala:
@rschmukler made the if statement changes, I also noticed a few other things that could be done differently, let me know what you think and I'll squish when you are okay with the changes.

ilanbiala:
@rschmukler [the test](https://travis-ci.org/rschmukler/agenda/jobs/46156590#L553) seems to have failed due to a timeout, not a code error.

ilanbiala:
@rschmukler did you have a chance to look at this?

rschmukler:
@ilanbiala sorry on the delay - merged. Thanks for the help!

ilanbiala:
@rschmukler no problem. If you are still making changes every so often, I would recommend using a code beautifier and linter in your editor of choice so that you have a consistent style.

ilanbiala:
@rschmukler are you open to a PR that adds the proper configuration files to the repository?

PR #136, opened 2015-01-12
Add .editorconfig

ilanbiala:
This will add some formatting consistency for the future. @rschmukler this is just a template. Please suggest changes that should be made and I'll fix up the PR so you can merge it.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,33 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+
+# Matches multiple files with brace expansion notation
+# Set default charset
+[*.{js,py}]
+charset = utf-8
+
+# 4 space indentation
+[*.py]
+indent_style = space
+indent_size = 4
+
+# Tab indentation (no size specified)
+[*.js]
+indent_style = tab
+
+# Indentation override for all JS under lib directory
+[lib/**.js]
+indent_style = space
+indent_size = 2
+
+# Matches the exact files either package.json or .travis.yml
+[{package.json,.travis.yml}]
+indent_style = space
+indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -3,20 +3,11 @@
 # top-most EditorConfig file
 root = true
 
-# Unix-style newlines with a newline ending every file
+# Unix-style newlines with a newline ending every file, default charset
 [*]
 end_of_line = lf
 insert_final_newline = true
-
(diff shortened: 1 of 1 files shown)


ilanbiala:
@rschmukler what changes should I make so we can merge this?

ilanbiala:
@rschmukler what else should I change in terms of the rules so that we can merge this?

rschmukler:
@ilanbiala I'm not sure this is a necessary file. I don't even use an .editorconfig. I realize that coding conventions are a good thing, but really I think this might be overkill. The styling conventions should be mostly obvious from the files, and adding a plugin dependency doesn't seem worth polluting the repo.

I'd be interested to hear what other people think - any thoughts @droppedoncaprica @nwkeeley @bars3s?

ilanbiala:
@rschmukler .editorconfig files are usually used in combination with an editor plugin/package that reads the file and applies the formatting, so for people who aren't as consistent with their files, at least there won't constantly be commits changing the ending newline around or silly things like that.

nwkeeley:
@rschmukler @ilanbiala I've never used an .editorconfig. I just try and match the style of the code base I am working in. By default I try to stick with the style defined here: https://github.com/RisingStack/node-style-guide.

I guess a .editorconfig could be a good idea... but we can't really force anyone to abide by it...

ilanbiala:
@nwkeeley we can't but when working on a codebase with multiple developers, consistency gets kicked to the curb.

droppedoncaprica:
Personally, I'm not specifically sold on the effectiveness of an `.editorconfig` file.  If anything I think a `.jshint` or `.jslint` file would be more suitable due to more JS-specific rules being enforced (albeit this config you've supplied also dictates things such as tab indentation amounts and cross-platform specific things which aren't enforced by JSHint/ JSLint).

That being said, I don't think code formatting is *that* huge of an issue (for the most part).  It's something to keep in mind of when going over PR requests, but I'd think looking at the already established code would be enough to gleam what the standards are.  Not having a "spec" to adhere to is a hurdle in enforcing this, but again I don't see this as super-important.  Just my opinion though, I don't have strong feelings either way.

Moving on to the PR itself, any reason this PR has `.py` rules?

ilanbiala:
@droppedoncaprica it wasn't clear to me if there are any preferences, so I just left the options open?

droppedoncaprica:
@ilanbiala It's not a huge deal, just something I'd noticed.

ilanbiala:
@droppedoncaprica I do plan to remove it, but I just want to make sure that everyone knows what's possible.

rschmukler:
I'm going to close this, as I don't see much buy in.

ilanbiala:
@rschmukler what about a .jslintrc or something like that?

PR #141, opened 2015-01-28
scoped ignoreErrors variable

bompus:
Without this variable scoped, ignoreErrors leaked into the global scope.


bompus:
Any issue with getting this pulled in? I have a test failing right now because of the variable leaking into the global scope, which I don't think is the intention.

rschmukler:
So sorry, I just fixed this randomly the other day and published it. Thank you for opening the PR! It's fixed in the latest version of agenda

PR #151, opened 2015-02-11
Fix for when _findAndLockNextJob returns multiple jobs.

### LINKED ISSUE #131: Restarting the server may stop the recurring jobs forever
I'm starting my job every 30 minutes, but sometimes when restarting the server, the job stops working forever until I remove it from the `agendaJobs` collection manually, & I tried smaller periods for testing like 10 seconds, & it is now a consistent behavior, my code is like this:

``` coffeescript
agenda.start()
graceful = ()->
    agenda.stop ()->
        process.exit 0

process.on 'SIGTERM', graceful
process.on 'SIGINT' , graceful

agenda.define 'my job', my_job_function
agenda.every '10 seconds', 'my job'

```

And as a workaround, I decided to cancel all recurring jobs when stopping the server like this:

``` coffeescript
graceful = ()->
    agenda.cancel repeatInterval: { $exists: true, $ne: null }, (err, numRemoved)->
        agenda.stop ()->
            process.exit 0
```

With this workaround, now restarting the server doesn't affect any of the recurring jobs,

**but, I don't know why I have to cancel all the recurring jobs in order to be safe from such a situation?!**

droppedoncaprica:
After working out how Agenda pulls in jobs for processing, I noticed
that the _findAndLockNextJob method could return multiple jobs, which
ends up breaking jobs with the same definition.  The query grabs and
updates the list of jobs whose lock time is passed or should be ran,
meaning if you have multiple of the same jobs running and shut
down the server, it will end up never running those jobs again.

I do need to pick @rschmukler's brain about how to properly get tests
for these methods, as it seems that the methods in question are
private and unreachable using Mocha.

Fixes #131 

I'm mostly creating this PR right now to run tests, as I'm pushing this from a computer without Mocha / MongoDB installed.  Any ideas on how to get proper tests done for this issue @rschmukler / @nwkeeley?

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -20,7 +20,7 @@ var Agenda = module.exports = function(config) {
   if (config.db) {
     this.database(config.db.address, config.db.collection, config.db.options);
   } else if (config.mongo) {
-    this._db =  config.mongo;
+    this._db = config.mongo;
   }
 };
 
@@ -183,7 +183,7 @@ Agenda.prototype.saveJob = function(job, cb) {
   delete props.unique;
 
   props.lastModifiedBy = this._name;
-  
+
   var now = new Date(),
       protect = {},
       update = { $set: props };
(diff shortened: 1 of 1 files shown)


rschmukler:
Merging this in immediately. The tests for it will take some refactoring, I imagine.

droppedoncaprica:
Sorry about the formatting changes as well, I noticed a couple things that were off and figured I would take care of them with this PR.  I've got a couple tests on my local machine that I'll have a PR for in a few days.

Issue #144, opened 2015-02-01
Pause Specific Job


jgallen23:
Lets say I have a specific job that I've found a bug in.  Is there any way to pause just a specific job and just queue up all requests?  If not is that something you'd take a pull request on?

droppedoncaprica:
I believe you can do this by calling `job.disable()` on the job (and `job.enable()` to re-enable it again).  You'll also have to call `job.save( cb )` to ensure it gets set in the database as well.  [Source](https://github.com/rschmukler/agenda/blob/master/lib/job.js#L115)

Alternatively you should be able to find the job in the DB and manually update the record to have `disabled` set to true, and it shouldn't run.

Let me know if that doesn't work!

jgallen23:
that is on a specific job though, right?  I'd like to be able to do it across the board for future jobs as well.  Something like `agenda.pause('email-notification')` and then agenda will just queue those up instead of actually processing them.

droppedoncaprica:
Yeah, that would be on specific jobs.  You can wholesale disable Agenda using `agenda.stop()`, but that won't be on a per-job basis.

jgallen23:
it looks like `job.disable()` and then `job.save()` actually runs the job.  Is that a bug?

droppedoncaprica:
Well... bug-ish.  

The `job.pause()` method isn't discussed in the README at all, though it is present in the actual code.  If you would like to dive in and see where Agenda isn't respecting the `disabled` attribute when looking up jobs, we can get a PR in.

As of right now though, my time.  It hurts. :frowning:

Nishchit14:
@droppedoncaprica 
`job.pause()` is not present there, firing error of undefined.

Now for pause I've did first `job.disable()` then `job.enable()` on un-pause. But after enabling job, It's not running (I've setup it with every 20 seconds)

After enabling data looks like this
```
    "priority" : 0,
    "nextRunAt" : ISODate("2017-01-12T07:03:32.718Z"),
    "repeatInterval" : "20 seconds",
    "repeatTimezone" : "Asia/Kolkata",
    "lastModifiedBy" : null,
    "lockedAt" : ISODate("2017-01-12T07:04:20.985Z"),
    "lastRunAt" : ISODate("2017-01-12T07:03:12.718Z"),
    "lastFinishedAt" : ISODate("2017-01-12T07:03:12.982Z"),
    "disabled" : false
```

PriyaJainDev:
[...]

PR #161, opened 2015-03-26
Check if job is disabled before running

dawnerd:
When using `.every()` and `.disable()` `disabled` isn’t checked when
the job is saved, resulting in the job running.

Related to issue #144


droppedoncaprica:
@rschmukler Hope you didn't mind me merging this in.  Do you want me to make a new tag and change log entry as well?

PR #162, opened 2015-03-27
updated to allow querying for a single job

brock:
Tests check for a single job by id and by name.


loris:
Not sure if you still need this a year later, anyway it would need to be rewritten (now using `mongodb-native` instead of `mongoskin`)

PR #167, opened 2015-04-07
SVG code climate and coverage badges

sahat:


droppedoncaprica:
Good job Travis, that didn't take long at all!  

In all seriousness, I'll probably merge in this PR soon.  I need to bump the version for a bugfix that was implemented anyway, just a matter of getting the time.

Issue #201, opened 2015-09-22
Peer dependency issue with npm3. Mongoskin needs to be updated.


manikandants:
```
npm WARN EPEERINVALID mongoskin@1.4.13 requires a peer of mongodb@~1.4 but none was installed.
npm ERR! Linux 3.19.0-28-generic
npm ERR! argv "node" "/usr/bin/npm" "install"
npm ERR! node v0.12.7
npm ERR! npm  v3.3.3
npm ERR! path /home/user/project/node_modules/agenda/node_modules/mongodb/node_modules/readable-stream/node_modules/core-util-is
npm ERR! code ENOENT
npm ERR! errno -2

npm ERR! enoent ENOENT, rename '/home/user/project/node_modules/agenda/node_modules/mongodb/node_modules/readable-stream/node_modules/core-util-is'
npm ERR! enoent This is most likely not a problem with npm itself
npm ERR! enoent and is related to npm not being able to find a file.
npm ERR! enoent 

npm ERR! Please include the following file with any support request:
npm ERR!     /home/user/project/npm-debug.log
```

mahnunchik:
+1

manikandants:
:+1:

nha:
Same with `npm 3.3.6` on osx.

npm WARN EPEERINVALID mongoskin@1.4.13 requires a peer of mongodb@~1.4 but none was installed.

What is the meaning of this warning ?

wesley-harding:
:+1:

loris:
Can be closed, mongoskin has been replace in `0.7.0`

PR #203, opened 2015-09-23
Set mongodb as direct dependency

omakoleg:


omakoleg:
Should fix issue with peerDependency. Similar to issue #201

Issue #211, opened 2015-10-15
Support updated mongoose v4.x with updated mongodb driver v2.x


sgilroy:
In my app we use agenda by providing the mongodb collection from mongoose. Something like this:
```
var mongoose = require('mongoose');
var Agenda = require('agenda');
var agenda = new Agenda({ mongo: mongoose.connection.collection('scheduledjobs') });
```

After upgrading to mongoose v4.1.10, I was getting the error `TypeError: Cannot read property 'concurrency' of undefined` because db.findAndModify passes a result to the callback that looks like `{ ok: 1, value: { _id: 'some-id', name: 'some-name' } }` instead of the expected `{ _id: 'some-id', name: 'some-name' }`.

This can be resolved by using the `res.value` instead of `res`.

Looks like this may be the underlying cause of the error reported in #163.

wesley-harding:
:+1: 

I'd really like to see this included.  I've updated mongo/mongoose and I've had to switch to pulling in sgilroy's branch.

PR #212, opened 2015-10-15
Handle updated response from findAndModify with mongodb driver v2.x

sgilroy:
For #211

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -208,6 +208,9 @@ Agenda.prototype.saveJob = function(job, cb) {
   }
 
   function processDbResult(err, res) {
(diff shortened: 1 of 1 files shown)


jdiamond:
Changes in this PR don't seem necessary anymore after the refactoring in this commit:

https://github.com/rschmukler/agenda/commit/4246c3bded7864a6896a771f4dacb5b8edee3635

Main difference is this PR made it possible to use both 1.x and 2.x of the MongoDB driver, but after the refactoring there's no support for 1.x in other places so I think this PR can be closed.

rschmukler:
Closing PR since we now use 2.x

PR #216, opened 2015-10-21
Rollback job completion callback to pre-0.7.0

loris:
Fix another breaking change brought by release 0.7.0 (https://github.com/rschmukler/agenda/commit/4246c3bded7864a6896a771f4dacb5b8edee3635)
Besides switching to mongodb-native driver, the commit also change how completed jobs are handled in a very opinionated way: All non recurrent jobs (failed ones too) are now removed from the database.
That change might be a good idea, we should debate on it, but it needs to wait a new major release.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -1,7 +1,7 @@
-/** @package 
+/** @package
 
     job.js
-    
+
     21/04/2015 NF   Removed database code. ie. From Job.remove()
 
   Last change: NF 21/04/2015 1:20:47 PM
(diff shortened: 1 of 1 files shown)


loris:
In my opinion, Agenda should not automatically remove completed jobs. it’s up to the user to hook into events (`complete`, `success` or `fail`) and decide what to do with the jobs: remove successful ones, retry failed ones caused by temporary errors (like network issues), notify alert system, etc.

PR #219, opened 2015-10-27
Emit ready event when using config.mongo

loris:
`config.mongo` is doing synchronous work, but added ready event for being consistent with `config.db` behavior (https://github.com/rschmukler/agenda/issues/213#issuecomment-151317062)

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -51,6 +51,7 @@ var Agenda = module.exports = function(config, cb) {
   } else if (config.mongo) {
     this._mdb = config.mongo;
     this.db_init( config.db ? config.db.collection : undefined );    // NF 20/04/2015
+    this.emit('ready');
   }
 };
 
--- a/lib/job.js
+++ b/lib/job.js
@@ -152,6 +152,7 @@ Job.prototype.fail = function(reason) {
     reason = reason.message;
   }
   this.attrs.failReason = reason;
+  this.attrs.failCount = (this.attrs.failCount || 0) + 1;
   this.attrs.failedAt = new Date();
   return this;
 };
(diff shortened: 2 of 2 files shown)


jdiamond:
Testing this PR. Did you mean to add the "failCount" commit? Doesn't seem related to the ready event feature.

loris:
No actuall,y it was meant for another PR but both ended up on my fork
master branch. I will fix that as soon as I get a computer.
On Tue, 10 Nov 2015 at 23:00, Jason Diamond <notifications@github.com>
wrote:

> Testing this PR. Did you mean to add the "failCount" commit? Doesn't seem
> related to the ready event feature.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rschmukler/agenda/pull/219#issuecomment-155581377>.
>
-- 
À plus tard,
Loris

--
co-founder @gowento
cell: (+33) 6 71 31 27 70
twitter: @loris
http://www.lorisguignard.com

jdiamond:
So this fix doesn't really work because it emits the ready event before the `Agenda` constructor has a chance to return. Since I can't call `.on('ready')` until the constructor returns, I miss it.

I think moving the emitting of the ready event to `db_init()` (after the indexes are created) might be better so that it only has to be in one spot. As it is now, it's possible to not pass in a config object to the constructor and use the `database()` or `mongo()` methods to set those options. They end up calling `db_init()`, but don't emit the ready event.

jdiamond:
I took a crack at this here:

https://github.com/jdiamond/agenda/tree/ready-event

Can submit a PR if you think it's a good approach.

loris:
@jdiamond You right, this is the best approach, I will close this PR and test yours on my setup.
Thanks!

Pyrolistical:
is there a workaround before this is fixed?

rschmukler:
@jdiamond please submit a PR upstream and I'll merge in!

PR #217, opened 2015-10-22
0.6.29

coty-crg:
I pulled from https://github.com/rschmukler/agenda/releases/tag/0.6.28 and branched off, made some fixes in my own branch and called it version 0.6.29.

This pull request would add in callbacks to the Agenda.prototype.schedule function, so large scale applications (like the one I was working on) can know when the mongo operations required by the agenda module are complete before continuing logic.

PR #224, opened 2015-11-07
Support success callbacks

mgregson:
Extending #217 to provide callbacks for more scheduling functions. Also correcting an error in MongoDB response handling preventing user-provided callbacks from being called when the MongoDB operation encounters an error.


loris:
Great addition!

lib/agenda.js
```diff
@@ -200,33 +200,33 @@ Agenda.prototype.every = function(interval, names, data) {
...
-    job.save();
+    job.save(cb);
     return job;
   }
 
-  function createJobs(when, names, data) {
+  function createJobs(when, names, data, cb) {
     return names.map(function (name) {
-      return createJob(when, name, data);
+      return createJob(when, name, data, cb);
```
Pyrolistical: this is a bug and will cause cb to be called multiple times.

need to use `async.map` or similar

Pyrolistical:
I need this, these methods should have never been implemented without callbacks or promises

Pyrolistical:
Thank you @rschmukler for the quick release

Zertz:
This is genius, thank you @Pyrolistical

lib/agenda.js
mgregson: You're right. I think this bug might be present in the code that mede it onto mainline too. Can you confirm?

rschmukler: I fixed this before merging it into master. Thank you @mgregson

Issue #226, opened 2015-11-14
.unique is not really unique


aef-:
The way it's implemented is as an alias to findAndModify with upsert enabled. So when I try something like:

``` js
jobs
.create('unique job', {userId: '100'} )
.unique({'data.userId': '100'})
.schedule('now')
.repeatEvery("1 day")
```

The job will keep updating its schedule if I run that multiple times when I just want it to only update the DB on creation. #105 also chimes on this.

I propose we default to using setOnInsert for `unique`, unique shouldn't change anything, unless I've missed something through all of this.

Thanks!

PR #227, opened 2015-11-14
Add option to unique that prevents update to job when already exists

aef-:
Adds ability to set `insertOnly` to `.unique` that uses setOnInsert instead of set for properties.
#226


OmgImAlexis:
Closes https://github.com/agenda/agenda/issues/226

PR #235, opened 2015-11-23
Fix repository URL at NPM

luanmuniz:


omakoleg:
Also was surprised about wrong link

rschmukler:
Don't know how that snuck in there. Thanks :)

Issue #104, opened 2014-10-09
Still getting several "ClientCursor::staticYield can't unlock b/c of recursive lock" warnings


brianmarco:
With regards to the following issue:

https://github.com/rschmukler/agenda/issues/82

We're still getting the "ClientCursor::staticYield can't unlock b/c of recursive lock" warning several times in our log when we get to the `_findAndLockNextJob` function.  We get so many of them that our log grows by GB's in minutes and the other services on the box crawl to a halt.  

For context we're running agenda with over 150k jobs.

From what we've been able to read up on it would seem that there still isn't the appropriate index needed for the `findAndModify` query being fired in this function.

Some other feedback we got was that the `findAndModify` query isn't the best approach, and that instead an `update` and then `find` would scale better.

Any chance anyone here is having the same problem and looking into a solution?

bedrierol:
I'm experiencing this problem @ windows MongoDB 2.6.6

bedrierol:
I've added a limit(, 'limit': 1) to findAndModify function @ _findAndLockNextJob @ Agenda and seems fine at high jobs..

nwkeeley:
I experienced this same issue. GB's of recursive lock warnings when our agenda collection exceeded 1,000,000 documents. 

I forked the repository and modified the findAndLock next job query and also altered the ensureIndex's in agenda.js.

We are now able to scan over a collection containing 10,000,000+ jobs with an occasional recursive lock log entry here and there. But the performance improvement was night/day. 

If anyone is curious I'll post what I did in my fork

alemhnan:
@nwkeeley +10. I will be very very interested!

nwkeeley:
I am not sure how others are using Agenda but everything was running great performance wise until we hit 500k-1 million jobs. After that happened we started to notice serious recursive lock issues (GB's of logs within hours...) 

We are still experimenting with performance issues but here is what we are currently doing... the tweaks below have allowed us to go from 1 million jobs (where we hit the recursive lock issue) to 10+ million, haven't really pushed it further than that since our high availability environment hasn't been finalized and we were testing on a single server development environment.

1. Removed the individual ensureIndexes() in lib/agenda.js. Replaced with a composite index (remember order matters!):  
this._db.ensureIndex({name: 1,  disabled: 1, nextRunAt: 1, lockedAt: 1}, ignoreErrors);

[...]

PR #241, opened 2015-12-03
Modify query conditions in _findAndLockNextJob to better utilize inde…

sshen81:
Mongo was still subject to recursive locks under heavy usage - issue #104 

This modification resolves it by modifying the findAndLockNextJob query conditions so that common criteria are no longer subject to $or condition.


karamfil:
https://github.com/rschmukler/agenda/pull/300

This pull request gets rid of the or statement and is really utilising the indices also making the query/code simpler.

simison:
Closing this as stalled. Our codebase is very different these days. Thanks for contributing, though!

PR #245, opened 2015-12-07
Fixes ReferenceError: MongoError is not defined

### LINKED ISSUE #244: ReferenceError: MongoError is not defined
When the db server goes down, is restarted, etc, agenda crashes the process because of the following:

ReferenceError: MongoError is not defined
    at Agenda._findAndLockNextJob (/node_modules/agenda/lib/agenda.js:381:13)

bompus:
Fixes #244


fabien:
@rschmukler this should definitely be fixed. It's not only long overdue, but prevents us from properly debugging our setup, because the actual error won't show up. Could you please have a look?

PR #246, opened 2015-12-12
Child job execution with next support

seanfuture:
Needed this for one of my projects.  Haven't had time to update the README with how to use, examples, etc., but a small snippet is listed below.  Thanks for your work on Agenda!

```
jobs.setup = function( job, done ) {

    async.parallel( [
        self.queue.child( 'setup_ensureEnvironmentConfigured', 'download', job ),
        self.queue.child( 'setup_ensurePrerequisiteDataPresent', 'download', job )
    ], done )

}
```

In this example, "download" is executed when all child jobs are complete.  Performance impact is incurred only when a parentid and a next value on a job is present and, in that case, on completion of the child job it checks the DB to see if all other child jobs are finished ( other than itself ).

This is all "now" centric of course, but is useful when scheduling / currently executing a "graph" of related jobs with dependency mappings.  Can also simply provide a "next" job name for chaining if desired and not use child jobs, which is mainly useful only for segmenting tasks and/or dynamic chaining of jobs depending on conditions.

Thanks for your time

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -149,6 +149,26 @@ Agenda.prototype.jobs = function( query, cb ){
 };
 
 
+Agenda.prototype.childJobsComplete = function( parentid, completingid, cb ){
+  var self = this;
+  this._collection.findAndModify(
+    {
+      _id: parentid,
+      childJobsLastRun: { $exists: false }
+    },
+    [], // sort
+    {
+      $set: {
+        childJobsCompleteAt: new Date(),
+        childJobsLastRun: completingid
+      }
+    },
+    { new: false },
+    cb
+  );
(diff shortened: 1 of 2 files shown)


raistlinthewiz:
will this be merged?

loris:
IMO, that’s the kind of features that would makes a lot of sense as a plugin (to avoid bloating agenda core). Anyway, some tests and README.md doc/examples/use-cases would help it being merged.

seanfuture:
Certainly agree - Apologies as I've been quite busy and haven't had time to
augment as I'd like to

leonardlin:
Closing this because it's stale

PR #252, opened 2015-12-28
Improve job failure handling

loris:
- Add a `failCount` job property which refers to the number of time a job has failed
- Move down the `fail`/`success` events triggering inside the `save` callback, to be consistent with the `complete` event

The previous changes give the developer correct `failCount`, `failedAt` and `lastFinishedAt` values, and allow:
- Fetching of all the failed jobs using the following mongodb query:
`{failedAt:{$exists: true}, $where:'this.failedAt >= this.lastFinishedAt'}`
- Retrying of failed jobs due to temporary reasons. For example, the following snippet will retry any jobs failed due to timeout or network error, at most 5 times, using an exponential backoff strategy (thanks to https://github.com/tim-kos/node-retry):
```
agenda.on('fail', (err, job) => {
  if (job.attrs.nextRunAt === null) {
    if (/(timeout of \d+ms exceeded|read ECONNRESET|connect ECONNREFUSED)/.test(err.message)) {
      const timeouts = retry.timeouts({ minTimeout: 10000, factor: 10, retries: 5 });
      const timeout = timeouts[job.attrs.failCount - 1];
      if (timeout) {
        job.attrs.nextRunAt = moment().add(timeout, 'milliseconds').toDate();
        job.save();
      }
    }
  }
});
```

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -51,6 +51,7 @@ var Agenda = module.exports = function(config, cb) {
   } else if (config.mongo) {
     this._mdb = config.mongo;
     this.db_init( config.db ? config.db.collection : undefined );    // NF 20/04/2015
+    this.emit('ready');
   }
 };
 
--- a/lib/job.js
+++ b/lib/job.js
@@ -152,6 +152,7 @@ Job.prototype.fail = function(reason) {
     reason = reason.message;
   }
   this.attrs.failReason = reason;
+  this.attrs.failCount = (this.attrs.failCount || 0) + 1;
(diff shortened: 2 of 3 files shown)


wootwoot1234:
Should the failCount be reset when the job succeeds in repeating jobs?  Using your example above doesn't work on repeating jobs because the failCount is never reset after a failure so the next time it fails it starts with more than 0 failures.

wootwoot1234:
Submitted pull request here https://github.com/agenda/agenda/pull/732

PR #254, opened 2015-12-31
Respecting job priority for on the fly job lock and queueing.

jakeorr:
Putting @CamHenlin's snippet in a function and reusing it for the on-the-fly-lock case.


rschmukler:
Thanks for this!

jakeorr:
No prob, thanks for the quick merge!

PR #257, opened 2016-01-06
Fix agenda throwing an exception when starting a job defined on another instance

loris:
Starting from server A a job (using `agenda.now` for instance) defined on server B is throwing an error right now: the `saveJob`method is calling `processJobs` with the job passed as `extraJob` parameter, there is no check that this job is defined before locking and requesting its execution.


rschmukler:
Thank you @loris !

PR #250, opened 2015-12-17
Added A Fix To Stop Concurrent Jobs Executing Out Of Sequence

nialldonnellyfh:
Related to: https://github.com/rschmukler/agenda/issues/231

When using a concurrency of 1, it was possible for jobs to execute out of sequence.

A Job in memory with a `nextRunAt` time less than the `nextRunAt` time of a job loaded from the database would execute immediately causing a Job to run between intervals.

The test in this PR highlights the issue. The fix has not broken any existing tests.

This will fix the bug on the master branch. If possible can you also apply the patch to version 0.6.28 and publish as 0.6.29 as I am using MongoDB 2.4.

I have the patch at https://github.com/rschmukler/agenda/compare/0.6.28...nialldonnellyfh:fix_concurrency_bug

Kind Regards
Niall Donnelly

PR #258, opened 2016-01-08
do not block normal jobs from running concurrently

jdiamond:
This is a fix for a problem that was introduced with #250 which is preventing normal jobs from running concurrently. Without this, normal (not single) jobs with the same name can get "stuck" and never run. All I did was require the job type to be single before checking to see if it should be skipped. All tests still pass.


rschmukler:
Thanks @jdiamond !

Issue #256, opened 2016-01-05
Jobs not processing on multiple servers in parallel


krishna-koushik:
Hi,

I really like what agenda does and we are planning to replace our current scheduler service with it and we did a small POC on this. In our POC we started agenda in clustered mode and more than 1 server/machine. So multiple agenda server processes in 2 servers. we followed the suggested pattern in the documentation when starting up agenda server in clustered mode.
when we started testing we found that 

```javascript
var options = {
        processEvery: '10 seconds',
        maxConcurrency: 8,
        defaultConcurrency: 1,
        defaultLockLifetime: 30000
    };
var agenda = new Agenda(options);
agenda.database('localhost:27017/agenda_test', 'agenda_jobs');

agenda.define('hello world', {lockLifetime: 30000}, function(job, done) {
                console.log('Hello World', 'Agenda.name', agenda._name);
                setTimeout(function () {
                    console.log('Done working...');
                    done();
                }, 25000);
    });
```

In another js file

```javascript
var agenda = new Agenda();
agenda.database('localhost:27017/agenda_test', 'agenda_jobs');
for(var i = 0; i<= 100; i++) {
agenda.now('hello world', {id: i});
}
```

Scenarios:
Case 1: an Agenda Server, clustered mode running on 8 processors (8 agenda workers)

Observed: Everything works expected and its great.

Case 2: more than one machine with Agenda Server in clustered mode running (8+ processes)

Observed: 
* Agenda Server on machine A locks all 100 jobs and machine B is in a waiting mode
* After 30,000 ms machine B starts to pick up some of the jobs which machine A did not finish and both machine A and machine B runs same jobs.

What am I missing here? i was under the impression that the agenda processes would processes 1 job or up to 16 jobs at a time from this configuration without conflict (all run in parallel without fighting over locks).

Update: I have updated the code part here for more information.

krishna-koushik:
Just checking in if someone else ran into this problem. Also to add additional information on the above problem here is the job object which ran as duplicates in 2 different servers.

In Machine 1
```javascript
{ agenda: 
   { _name: 'SchedulerServer55298',
     _processEvery: 10000,
     _defaultConcurrency: 1,
     _maxConcurrency: 8,
     _definitions: { hello world: [Object] },
     _runningJobs: [ [Circular] ],
     _jobQueue: [ [Object], [Object], [Object] ],
     _defaultLockLifetime: 30000,
     _events: 
      { 'success:hello world': [Object],
[...]

Issue #231, opened 2015-11-18
Issue Running Jobs In A Clustered Environment


nialldonnellyfh:
Hi, I really like Agenda and am starting to use it in a project. I am having an issue when running Agenda in a clustered node environment.

I would like to create a PR to solve this issue if it is a bug and would like to get your thoughts on it first.

Scenario: [Example Code To Reproduce](https://gist.github.com/nialldonnellyfh/1e826c4340506f660a50)
1. A node master process forking a number of worker processes.
2. Each worker process sets up the Agenda with a single job **testjob**. Each Agenda is set up for a **defaultConcurrency** of 1.
3. The **testjob** Job finishes in _1 second_ and is scheduled to run every _2 seconds_.
4. I expect that the **testjob** Job should execute every _2 seconds_.

[Example Output](https://gist.github.com/nialldonnellyfh/93fbb3420236ad7394e8)

From the output above, you can see process _6713_ starts executing the job normally every 2 seconds as expected. However, at [this](https://gist.github.com/nialldonnellyfh/93fbb3420236ad7394e8#file-example-output-of-test-job-L10-L11) point, both processes end up executing the job at the exact same time.

I am debugging the issue at the moment to find a solution and would be very interested in any comments that you have on the above scenario that I have missed.

I will update this ticket with any issues I can find from debugging.

Kind Regards
Niall Donnelly

nialldonnellyfh:
The issue seems to be related to more than a clustered environment.

I have added a unit test (https://github.com/nialldonnellyfh/agenda/tree/fix_concurrency_bug) to test concurrency that can reproduce the bug.

From debugging the code, it seems that the bug may occur between a Job saving and a deferred job executing on a timeout.

lfryc:
I've updated original test case with inter-process messages that makes master to print current number of concurrent job executions as well as the number of miliseconds from the start of last job execution, and after few seconds we are getting interesting output:

````
#testjob Finished Mon Dec 07 2015 10:40:18 GMT+0100 (CET) 59786
#testjob Starting Mon Dec 07 2015 10:40:18 GMT+0100 (CET) 59786
Last job started before 2005 ms
Concurrent job executions: 1
#testjob Finished Mon Dec 07 2015 10:40:20 GMT+0100 (CET) 59786
#testjob Starting Mon Dec 07 2015 10:40:20 GMT+0100 (CET) 59786
Last job started before 1522 ms
Concurrent job executions: 1
[...]

PR #260, opened 2016-01-12
Expired jobs

jdiamond:
This PR stops Agenda from running jobs that have expired after they were enqueued, but before they actually got a chance to start.

It's easy to recreate this by setting concurrency to 1 and scheduling multiple jobs that take close to their lock lifetimes to complete. Agenda will lock and enqueue all available jobs. Since it could only run 1 at a time and running that 1 job takes a while, the other jobs' locks would expire while waiting in the in-memory queue. Other Agenda instances (or even the current instance) would lock the job and run it.

This PR removes the fix from PR #250. Now that I understand the code better, I don't think it was correct. The fix in this PR is more general and doesn't have the problem I tried to address in PR #258.

This should fix issues #256 and #231.

This PR also includes changes to the tests so that each test uses its own instance of Agenda. Certain tests were randomly failing every time I ran the whole suite. I was eventually able to tell that jobs from previous tests were raising events after the next test started, confusing the next test with unexpected events. Using a new instance of Agenda makes the tests run slower, but I think the consistency is worth it.

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -515,12 +515,11 @@ function processJobs(extraJob) {
         if (jobDefinition.concurrency > jobDefinition.running &&
           self._runningJobs.length < self._maxConcurrency) {
 
-          if (job.attrs.type === 'single' && jobDefinition._lastCompletedJob) {
-            // If The Next Run Time At Is Less Than The Last Completed Jobs Next Run, Then Don't Run It Again.
-            if (job.attrs.nextRunAt < jobDefinition._lastCompletedJob.attrs.nextRunAt) {
-              jobProcessing();
-              return;
-            }
+          var lockDeadline = new Date(Date.now() - jobDefinition.lockLifetime);
+
+          if (job.attrs.lockedAt < lockDeadline) {
+            jobProcessing();
+            return;
           }
 
           self._runningJobs.push(job);
@@ -540,9 +539,6 @@ function processJobs(extraJob) {
 
     self._runningJobs.splice(self._runningJobs.indexOf(job), 1);
     definitions[name].running--;
-
-    //Assigning the last completed job for a specific job.
-    definitions[name]._lastCompletedJob = job;
 
     jobProcessing();
   }
--- a/test/agenda.js
+++ b/test/agenda.js
@@ -428,60 +428,6 @@ describe("agenda", function() {
     });
   });
 
-  describe('Scheduling Concurrent Jobs', function() {
-    it('do not run more than once in a scheduled time interval', function(done) {
-      var jobs = new Agenda({
-        defaultConcurrency: 1,
-        db: {
-          address: mongoCfg
-        }
-      });
-      var jobRunInterval = 400;
-      var jobRunTime = 200;
-      var processEvery = 300;
-      var successTime = 5000;
-      var lastRunAt;
-
-      function jobProcessor(job, done){
-        if(lastRunAt){
-          var timeSinceLastRun = new Date() - lastRunAt;
-
-          if(timeSinceLastRun < jobRunInterval - 50 || timeSinceLastRun > jobRunInterval + 50){
-            throw "INVALID Job Execution Time " + timeSinceLastRun;
-          }
-        }
-
-        lastRunAt = new Date();
-
-        setTimeout(function(){
-          done();
-        }, jobRunTime);
-      }
-
-
-      jobs.define('concjob', {concurrency: 1}, jobProcessor);
-      var interval = setInterval(function(){
-        clearInterval(interval);
-        jobs.stop(done);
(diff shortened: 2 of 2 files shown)


loris:
Great work! will try/deploy it in my apps soon.
Another use-case I had I think it will fix (that was brought by #250):
After completion I need some of my jobs to be run again later, so I update the `nextRunAt` attributes and save the job. The issue is that if others jobs where enqueued in between they would be locked instead of being run because of this:
```
if (job.attrs.nextRunAt < jobDefinition._lastCompletedJob.attrs.nextRunAt) {
  jobProcessing();
  return;
}
```

jdiamond:
@loris: I specifically had to add the call to `jobProcessing` because jobs were getting "stuck" and not running. Maybe that's the same thing you're referring to?

I'll look into that test failure in Travis later today. Surprised to see it. =(

jakeorr:
@jdiamond Thanks for working on this fix! It should be a very useful addition. I just did some quick testing and noticed something.

I'm using this job:
```javascript
agenda.define('test job', {
  concurrency: 1,
  lockLifetime: 2000
}, function (job, done) {
  console.log('Starting test job', job.attrs.data.i);
  setTimeout(function () {
    console.log('Completing test job', job.attrs.data.i);
    _.each(agenda._jobQueue, function (queuedJob) {
      console.log('still queued', queuedJob.attrs.data.i);
    })
    done();
  }, 5000);
});
```

And scheduling two of them with:
```javascript
agenda.now('test job', {
  i: 1
});
agenda.now('test job', {
  i: 2
});
```

Before the fix, the first job would run once and the second would run twice. After your fix, both jobs only run a single time (great!). What I noticed is that while job 2 is only being run once, it is still being queued twice. After job 2 completes, the second version of it is still queued although it does not run.

jdiamond:
Hi @jakeorr: I saw the same behavior. It looks weird, but I think it's OK.

Your example is a little strange, though. Since you're making your job take 5 seconds to complete but only locking for 2 seconds, isn't that guaranteeing the lock will expire and be eligible for running by other Agenda instances? I'm surprised job 1 didn't get run multiple times. I will have to check the code to see why.

The general behavior I observed is this: When Agenda ticks, all jobs that run soon get locked and enqueued. If locks on jobs expire while still in the queue, next time Agenda ticks, it will re-lock the jobs with the newly expired locks so now they're in the queue twice (possibly more). This is why I added the if statement to skip running the expired jobs.

So if we skip running the first instance of the job that expired while in the queue, I think this is OK even though it does feel odd to have two instances of the same job in the same queue.

We could set a timeout to purge the queue of expired jobs, but adding the if statement when trying to run the job seemed less complicated.

Is there a disadvantage to not purging the expired items from the queue as soon as they expire? Or some advantage to doing so? I guess clarity is one advantage.

But I have no idea how we could guarantee a job doesn't get run more than once if it takes longer to complete than its lock lifetime allows. Maybe jobs could use some API we create to extend the lock while it's running, but I think an expired lock basically indicates the job crashed so needs to be picked up to run again.

jakeorr:
Hey @jdiamond: Ah right, yes you are correct. I was missing the distinction between a job taking longer than its lock lifetime and a job being queued for longer than it's lock lifetime.

As a side note, [using `job.touch()`](https://github.com/rschmukler/agenda#touchcallback) would allow a job to run longer than its lock lifetime.

It doesn't look like expired jobs will stay in the queue for long. When `jobProcessing()` runs it pops the job off of the array and, if it fails the expiry check, isn't pushed back on. I corrected my test, added some console output into agenda, and am able to confirm that after my last job finishes `jobProcessing` runs again, picking up the duplicate job. It fails the expiry check and `jobProcessing` runs again finding an empty queue.

Agreed that you lose some clarify by having expired jobs hanging around in the queue until Agenda tries to run them. Also, think about this case: you have a job that takes 30 minutes to complete, your lock lifetime is 10 minutes, and concurrency is low (say concurrency=1). If you were to schedule 1000 jobs, the queue would be filled with duplicate jobs every 10 minutes and grow quite large before the expired jobs were removed.

What about including an expiry check in `enqueueJobs()` when we are spinning through the queue?

rschmukler:
Thanks a ton @jdiamond

nvartolomei:
There was a problem after #250.

I would have a job that should repeat every day. After it runs it was scheduled for next day. Then I create a new job that should be scheduled right now, because of the nextRunAt check in #250 this actually was ignored because lastJob.nextRunAt was greater than the job that is scheduled for right now.

Is this still the case with #260?

ps. any eta for next npm release?

loris:
@nvartolomei it is probably fixed with #260 (I had a similar problem: https://github.com/rschmukler/agenda/pull/260#issuecomment-170851418)

PR #265, opened 2016-01-14
Fixing bug where on-the-fly-locked jobs will not run

jakeorr:
During the on the fly lock of a job, `extraJob` was being queued without it's updated `lockedAt` time which resulted in the job being queued with a null lock time. Now that `lockedAt` is being used to determine the freshness of the job (from https://github.com/rschmukler/agenda/pull/260), this is a problem. When the run is attempted, the null locked time fails the expiry check, and the job is not run (but remains locked in the database). Using the updated job document after locking solves the problem.

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -445,9 +445,9 @@ function processJobs(extraJob) {
   } else if (definitions[extraJob.attrs.name]) {
     // On the fly lock a job
     var now = new Date();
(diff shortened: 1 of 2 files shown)


jakeorr:
**updated:** Test added

The issue I described occurs when a job that is already in the database is rescheduled and saved (like in the case of a retry). I added a failing test that this PR fixes so you can see an example of what I'm describing.

rschmukler:
Thanks @jakeorr definitely appreciate it.

Issue #79, opened 2014-06-26
New feature request - Consider TimeZone


sureshballa:
First of all I would like to thank for creating such a beautiful flawless scheduling functionality. 
I would like to see new functionality for running scheduler based on different timezone for each job. For example, one job should run as per EST timezone and other one should be based on PST timezone. This way, I will be expose this module as HTTP service and deploy any where across the world and still make usable for any timezone users.

Since cron module, a module agenda internally uses, supports timezone, in the computeNextRunAt function of job.js should pass timezone information from job request to CronTime constructor. This way, nextRunDate is computed as per target timezone that is mentioned for job.

It would be great if this request is considered or I am equally existed to implement this feature if I can contribute.

Thanks,
Suresh

keithrz:
So now that node-cron accepts a timezone parameter, could repeatEvery be changed to accept a timezone parameter?

loris:
I think it would make sense (and would be simple) to allow setting the timezone of the Agenda instance globally (which you probably wanna set to the timezone of the server defining the jobs).
For instance:
`var agenda = new Agenda({timezone: 'Europe/Paris'});`
Store/default it in the Agenda constructor:
`this._timezone = config.timezone || 'Europe/London';`
And passing the value to the CronTime instantiation:
`var cronTime = new CronTime(interval, this.agenda._timezone);`

keithrz:
I'd prefer setting the timezone for each job.  We have some clients on the East Coast, and some one the West Coast.  Say the West Coast client wants their jobs run at 12 noon, 3pm, 6pm, 5 minutes into each hour:
5 12,3,6 * * *
and the west coast client wants their jobs run at 12 noon, 3pm, 6pm, 9pm:
5 12,3,6,9 * * *

In order to implement this currently, we're using logic similar to what @droppedoncaprica described, which works great except for the shifts to/away from Daylight Saving Time, as @miira aluded to.  
Let's say our server is an AWS box, and local time is always in UTC.  So during DST, our jobs had repeatIntervals (in UTC hours) of:
5 19, 22, 1 * * *
and:
5 16, 19, 22 * * *

so once DST ended, we had to change the jobs to:
5 20, 23, 2 * * *
and:
5 17, 20, 23 * * *

and we also needed to set "nextRunAt", unless we modified the repeatInterval before the last job before the time shift.

[...]

PR #267, opened 2016-01-16
Adding repeatTimezone

jakeorr:
This feature allows `options` to be passed to `agenda.every()` and `job.repeatEvery()`. Currently, the only property of `options` is `timezone` which is used in conjunction with a cron string interval to determine the correct nextRunAt for the specified timezone.

for #79

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ agenda.define('delete old users', function(job, done) {
 
 agenda.on('ready', function() {
   agenda.every('3 minutes', 'delete old users');
-  
+
   // Alternatively, you could also do:
   agenda.every('*/3 * * * *', 'delete old users');
 
@@ -328,9 +328,9 @@ agenda.define('say hello', function(job) {
 
 ## Creating Jobs
 
-### every(interval, name, [data], [cb])
+### every(interval, name, [data, [options]], [cb])
 
-Runs job `name` at the given `interval`. Optionally, data can be passed in.
+Runs job `name` at the given `interval`. Optionally, data and options can be passed in.
 Every creates a job of type `single`, which means that it will only create one
 job in the database, even if that line is run multiple times. This lets you put
 it in a file that may get run multiple times, such as `webserver.js` which may
@@ -340,6 +340,9 @@ reboot from time to time.
 
 `data` is an optional argument that will be passed to the processing function
 under `job.attrs.data`.
+
+`options` is an optional argument that will be passed to `job.repeatEvery`. In order to use
+this argument, `data` must also be specified.
 
 `cb` is an optional callback function which will be called when the job has been
 persisted in the database.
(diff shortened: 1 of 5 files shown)


jakeorr:
Hmmm, that error...it didn't appear in my testing and I'm not sure what to make of it. @jdiamond is this what you were seeing when you submitted #260?

```
 1) agenda Locking does not process jobs with expired locks:
     Uncaught Error: expected 4 to equal 3
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
      at null._onTimeout (test/agenda.js:1255:35)
```

jdiamond:
Yes, this is the issue I was trying to fix in #260. Sadly, that error still occurs a small % of the time (was 100% before I added the check for expired jobs).

I've added a bunch of logging in my branch to try to track down where the (presumed) race condition is. I'll update when I think I've found it, but this failing test is definitely not related to the change you made in this PR.

By the way, I've got an open issue against my app that this PR will help me close so I'm happy to see it. :+1:

loris:
Awesome! Thanks a lot! Some feedbacks:
- now that `moment` is a dependency, we might want to remove `human-interval` and `date.js`
- Agenda constructor should accept a global `timezone` option

jakeorr:
@jdiamond: Ah okay, well that's good to know. Thanks for the info :) One thing I noticed when running the tests is that all of those agenda instances are creating their own mongo connections which are not being closed. Another PR that I have up [closes `jobs._mdb` after each test](https://github.com/rschmukler/agenda/pull/265/files#diff-65704758a2c4406e3c56f4ca06c2518cR66). I wonder if that has anything to do with this error.

I felt the same way about #260, great to see :+1:

@loris: I will add timezone as a global option and see about using it where appropriate. So far I'm seeing a place for it in `job.schedule` and `job.repeatAt`.

Replacing human-interval and date.js could be an option. It might be good to simplify the use of dates. As I understand it though, moment doesn't do everything that those modules do so it would mean changing functionality in a number of places. I think it would deserve its own issue/PR; I would rather not include it here.

jakeorr:
Hi @loris: I've run into a problem trying to incorporate timezone into job.schedule and job.repeatAt. Both use `date.js` to determine the nextRunAt date and the module does not support timezones. I think agenda might be better off leaving the timezone option specific to the cron string case until either date.js supports timezones or use of the module (date.js) is removed. I think a global timezone option would be misleading if agenda is only using it with cron strings in `every` and `repeatEvery`. Let me know what you think.

rschmukler:
@loris and @jakeorr my thought is to keep `date.js` and `human-interval`, so as to not break agenda's readable scheduling API. Relying on cron syntax for timezone's seems reasonable to me as it is a semi-advanced feature. 

Aditionally, @matthewmueller (author of `date`) is usually very good about merging in PRs so it might be possible to get this merged upstream. 

Also worth noting that only `moment-timezone` got added as dependency, not proper `moment`. `moment-timezone` is much smaller than `moment`. I think that it isn't worth so drastically changing the API to switch to moment.

Let me know your guys' thoughts.

loris:
Yes it makes sense to keep `date.js`! I thought all `moment` was included in `moment-timezone`

jakeorr:
@rschmukler agreed :+1:

PR #281, opened 2016-02-02
Fixing computeNextRunAt timezone bug.

jakeorr:
I noticed that my jobs scheduled with a repeatTimezone were getting scheduled for the wrong time. The issue was occurring in `Job.computeNextRunAt()`. If `cronTime._getNextDateFrom(lastRun)` returns the same date, `lastRun` is adjusted by adding 1 second and nextDate is recalculated. The date used to recalculate was not considering timezone resulting in the intermittent error I was seeing.

This PR fixes the issue and includes a test that fails with on the current version.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -71,15 +71,21 @@ Job.prototype.computeNextRunAt = function() {
   }
   return this;
 
+  function dateForTimezone (d) {
+    d = moment(d);
+    if(timezone) d.tz(timezone);
+    return d;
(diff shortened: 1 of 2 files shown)


shaunymca:
can this be merged!? I'm seeing this same error.

rschmukler:
Thanks @jakeorr !

jakeorr:
@rschmukler :+1: :+1:

PR #285, opened 2016-02-16
Use callback to handle MongoClient connection errors if we can.

inkadnb:
When using the Agenda constructor or .database, if a connection error occurred there was no way for me to catch the exception.

If a callback was is in, it will utilize that callback to pass back the error, otherwise it will throw like it has always been.


rschmukler:
Thank you @inkadnb, definitely useful!

PR #286, opened 2016-02-18
Implementing lock limit

jakeorr:
This addresses the issue of multiple machines/processes not sharing locks well described in #256 by providing a `lockLimit` setting. `lockLimit` controls the number of jobs an instance of agenda will have locked at any moment. It can also be specified on a job definition to control how many of a specific job can be locked at one time.

As part of this change I moved on-the-fly locking into a recursive method that waits for the lock response before attempted the lock of the next one. Previously many on-the-fly locks could be started before any responses were received from the database, making it difficult to control the number of locked jobs without introducing some tricky race conditions.

Some minor formatting changes are also included to bring some newer code in line with the existing code's style.

I also moved the "Locking" test up into the "job lock" test section where it seemed to be a better fit.

### DIFF
--- a/README.md
+++ b/README.md
@@ -230,7 +230,7 @@ var agenda = new Agenda({maxConcurrency: 20});
 
 ### defaultConcurrency(number)
 
-Takes a `number` which specifies the default number of a specific that can be running at
+Takes a `number` which specifies the default number of a specific job that can be running at
 any given moment. By default it is `5`.
 
 ```js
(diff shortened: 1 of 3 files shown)


krishna-koushik:
Awesome :)

rschmukler:
Fantastic!

loris:
@jakeorr @rschmukler I just found out that this merge introduced a bug (more info below). I will have a look at it, but @jakeorr you might have a better idea of what’s going on.

The bug is that if you create multiple jobs very quickly (let’s say 10 jobs at once), a random number of these 10 jobs will be ran multiple times instead of once.

Here is a simple snippet to repro:

```
agenda.on('ready', () => {
  agenda.start();

  for (var i = 0; i < 10; i++) {
    agenda.now('test-job');
  }
});

const ids = [];
agenda.define('test-job', (job, done) => {
  const id = job.attrs._id.toString();
  console.log(id, ids.indexOf(id) > - 1 ? 'WARNING' : '');
  ids.push(id);
  done();
});
```

The output:
```
56fabf6d00bd35df7fe9f275
56fabf6d00bd35df7fe9f279
56fabf6d00bd35df7fe9f276
56fabf6d00bd35df7fe9f27b
56fabf6d00bd35df7fe9f277
56fabf6d00bd35df7fe9f27d
56fabf6d00bd35df7fe9f278
56fabf6d00bd35df7fe9f27c
56fabf6d00bd35df7fe9f27a
56fabf6d00bd35df7fe9f27e
56fabf6d00bd35df7fe9f277 WARNING
56fabf6d00bd35df7fe9f276 WARNING
56fabf6d00bd35df7fe9f27a WARNING
56fabf6d00bd35df7fe9f275 WARNING
56fabf6d00bd35df7fe9f27e WARNING
```

jakeorr:
Hi @loris, good catch. Thanks for the repro steps, I am having a look.

jakeorr:
@loris: Okay, got it. This is occurring because on-the-fly locking now happens serially. In the case above, while a job was waiting to be locked on the fly, it was being queued by `jobQueueFilling` and completed before the on the fly lock was processed. The query for the on the fly lock was only checking for `lockedAt: null` which will be true if the job has been not been run or has already been completed. The PR I'm just about to create includes `nextRunAt` in the query.

PR #289, opened 2016-02-24
Add agendash as alternative ui

joeframbach:
Add agendash as alternative ui


rschmukler:
Thanks for this and your work on agendash!

PR #294, opened 2016-03-05
Update date and cron dependencies

markbahnman:


markbahnman:
It looks like the only place the tests fail is on [line 112 of lib/jobs.js](https://github.com/rschmukler/agenda/blob/cbf93c38fb1e486f7997168c1d2fa146caa478c9/lib/job.js#L112) where you're effectively comparing ```Date.now()``` to ```Date.now()``` which - depending on the timing - won't resolve to true. It looks like there might be plans to either throw an error or return null on an invalid string (mightmatthewmueller/date#12) but nothing is supported at the moment.

loris:
Closed by #311

PR #300, opened 2016-03-14
Optimize findAndModify

will123195:
Since findAndModify locks the db during search, the following optimizations should minimize the impact:
- Remove `$or` to prevent the optimizer from using the wrong index.
- Use `disabled: false` instead of `disabled: {$ne: true}` becasue `$ne` can't use a B-tree index.

This is a breaking change, so you may want to keep this in mind for version 1.0 to help prevent `ClientCursor::staticYield can't unlock b/c of recursive lock` errors.

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -39,6 +39,7 @@ var Agenda = module.exports = function(config, cb) {
   this._lockedJobs = [];
   this._jobQueue = [];
   this._defaultLockLifetime = config.defaultLockLifetime || 10 * 60 * 1000; //10 minute default lockLifetime
+  this._notLocked = new Date(1);
 
   this._isLockingOnTheFly = false;
   this._jobsToLock = [];
@@ -337,6 +338,8 @@ Agenda.prototype.saveJob = function(job, cb) {
     if (Object.keys(protect).length > 0) {
       update.$setOnInsert = protect;
     }
+    update.$setOnInsert = update.$setOnInsert || {};
+    update.$setOnInsert.lockedAt = this._notLocked;
     // Try an upsert.
     this._collection.findAndModify({name: props.name, type: 'single'}, {}, update, {upsert: true, new: true}, processDbResult);
   } else if (unique) {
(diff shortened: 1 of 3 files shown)


loris:
Looks great!
Do you have any benchmarks to share? Also I'm wondering about semver, it looks like it breaks backward compat (users expecting `lockedAt` to be `null` for non locked jobs)

michelem09:
+1 I think this should be merged asap, it solves a lot of problems with mongoDB performance.

simison:
@gastonelhordoy or @rschmukler got time to review this? It sounds like a major improvement.

michelem09:
Thanks @simison I did some tests and the changes decrease a lot (at least 20%) the resource needed by MongoDB.
Hope to see this released soon and if you need a hand, I'm here.

lushc:
After rebasing this branch and testing it I've noticed the following issues:

1. ed8b6ce fails the "runs a one-time job after its lock expires" test introduced in #306:

>  Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.

I don't think this can be addressed without [extracting the unlocking mechanism](https://github.com/rschmukler/agenda/issues/370#issuecomment-250419327) as proposed by @loris.

2. c678db7 *sometimes* fails the "should not run if job is disabled" test:

```
$ mocha --grep 'should not run if job is disabled'

  agenda
    Job
      Integration Tests
        .every()
          1) should not run if job is disabled


  0 passing (488ms)
  1 failing

  1) agenda Job Integration Tests .every() should not run if job is disabled:
     Uncaught Error: expected 26 to equal 0
```

Looking at the DB profiles for when the test [fails](https://gist.github.com/lushc/7fff72a63ce1bdd37dfcc07de74763b1#file-fail-profile-json) and [passes](https://gist.github.com/lushc/7fff72a63ce1bdd37dfcc07de74763b1#file-pass-profile-json), it seems that the first upsert sometimes contains the disabled property in the $set which conflicts with $setOnInsert, because you cannot update the same field in both operators. Even though this causes an exception in Mongo it silently fails (for me anyway on 2.6.10), which I guess puts things in a weird state and the job's disabled property is overwritten, so it starts running.

Guarding against this scenario seems to make the test pass consistently:
```js
if (!props.hasOwnProperty('disabled')) {
  update.$setOnInsert.disabled = disabled;
}
```

simison:
@will123195 would you have motivation to work again on this PR based on @lushc's comments?

OmgImAlexis:
@lushc this may or may not be related by I noticed quite a few of the tests fail between 5 and 10 percent of the time because of timeout issues. There seem to be a bunch of reoccurring tests that it happens to as well as some that I've only seen it happen to once or twice.

Maybe we need to look into that separately.

For that exact test I get a similar error with it ranging from 26 or 29 depending on how quickly I run the test suite back to back. Leaving 30s - 1m between test runs seems to allow it to pass or fail with 26 if I leave less than that I get upwards of 28.

simison:
Follow ups at duplicate #476

Issue #150, opened 2015-02-08
Job queue blocked in runOrRetry() 


plus-:
I have 2 type of jobs:

> agenda.define('parent', { priority: 'low', concurrency: 1 }, spawnChildren
> agenda.define('child', { priority: 'high', concurrency: 5 }, doneParentIfAllChildrenDone

Basically, parent job schedule several immediate children jobs. When all the childrens are completed, the done() callback from parent is invoked.

That's because I want all children to finished before starting another parent one (thanks to concurrency argument set to 1).

The issue is if another parent job is scheduled it blocks the jobProcessing because it keeps pushing back itself to the top of the queue. That's because the the first parent isn't completed (done not invoked), but then it block the processing of the remaining children from that first parent.

https://github.com/rschmukler/agenda/blob/master/lib/agenda.js#L393

That's because all job definitions share the same job queue. Should I run 2 instance of agenda in this case?

droppedoncaprica:
I'm honestly not sure if that way of doing it would work or not.  Since the primary job would be running with the concurrency being set to 1, I don't know if Agenda would kick off those children jobs until the main job is finished (and on top of that, have that single job sitting open and unfinished for so long).

That being said, I'm not super-familiar with how concurrency is implemented.  It's an interesting use-case though.  Any reason why you couldn't kick off the children processes yourself in the main job?

plus-:
That's a good question. Those children processes are intensive so I wanted to benefit from a job queue to control their execution.

droppedoncaprica:
Awesome, thanks for letting us know.  Since Agenda doesn't really have a concept of having parent processes (especially ones that manage other Agenda sub-tasks), I'm not sure how this could work on our end. ( @rschmukler may have a better idea here. )

jakeorr:
Good question @plus-. I'm seeing that the same blocking can happen even without the parent/child setup. I'm working with two jobs which are not dependant on one another:

```javascript
agenda.define('job1', { concurrency: 1 }, job1Def)
agenda.define('job2', { concurrency: 5 }, job2Def)
```

If I kick off two versions of job1 and then a single version of job2 (all scheduled to run immediately), the first job1 starts and the second job1 tries to start, but is pushed back onto the queue since concurrency is reached. Agenda doesn't get to job2 at all until job1's concurrency stops blocking.

[...]

PR #304, opened 2016-03-18
Stop concurrency from blocking other jobs

jakeorr:
Stops jobs that have hit their concurrency limit from blocking other jobs in the queue.

for #150


OmgImAlexis:
Closes https://github.com/agenda/agenda/issues/150

PR #305, opened 2016-03-19
throw on processJobResult error

imsnif:
Just a very small thing - this happened to me when the connection to Mongo dropped. Without this, the error I got was:
TypeError: Cannot read property 'attrs' of undefined

rather than:
MongoError: server localhost:27017 sockets closed


imsnif:
My bad - didn't notice the failing test. Will try to take a look at it later.

imsnif:
Fixed now, sorry about that.

PR #306, opened 2016-03-21
Picking up any job with an expired lock (not just recurring or queued).

### LINKED ISSUE #302: Running jobs with stale locks will not be restarted
From my understanding of the code, [this line](https://github.com/rschmukler/agenda/blob/cbf93c38fb1e486f7997168c1d2fa146caa478c9/lib/agenda.js#L419) from the `_findAndLockNextJob` method is how jobs with stale locks are picked back up and requeued. The issue is that the query will not match jobs where `nextRunAt` is null (ie. non-recurring jobs that are currently running). Based on my what the readme says, it seems to me that this is not the expected behaviour. Here are the points I am thinking about:

from https://github.com/rschmukler/agenda#defaultlocklifetimenumber
> The lock is useful if the job crashes or times out.

from https://github.com/rschmukler/agenda#multiple-job-processors
> This will ensure that no other job processor (this one included) attempts to run the job again for the next 10 seconds. If you have a particularly long running job, you will want to specify a longer lockLifetime.

also, the [`.touch()`](https://github.com/rschmukler/agenda#touchcallback) method implies that a running job's lock could expire. Currently, this would only be true for recurring jobs since they have `nextRunAt` set.

Please do let me know if I am misunderstanding something. Maybe someone has some insight that I have not considered.

Here is the some sample code as an example of the issue:

```javascript
agenda.define('test locking job', {
  lockLifetime: 1000
}, function (job, done) {
  console.log('starting', job.attrs);
  setTimeout(function () {
    console.log('finishing', job.attrs);
    done();
  }, 10000);
});

agenda.processEvery(2000);

agenda.now('test locking job');
```

The job starts, and then 10 seconds later it finishes. I would expect it's lock to expire during the wait time and for another instance of the job to be started.

jakeorr:
fixes #302 

Small change to agenda.js so that running locked jobs will be picked up again if their locks expire. I went ahead with picking up _any_ job with an expired lock. I haven't been able to think of a case where this would cause a problem. If anyone can come up with a case where this could be an issue, post a comment and I will take a look.

I added one test for one-time jobs being picked up after locks expire and I modified another test which didn't make sense anymore (formerly called `'does not process jobs with expired locks'` and now `'does not process locked jobs'`).

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -416,7 +416,7 @@ Agenda.prototype._findAndLockNextJob = function(jobName, definition, cb) {
         $or: [
           {name: jobName, lockedAt: null, nextRunAt: {$lte: this._nextScanAt}, disabled: { $ne: true }},
           {name: jobName, lockedAt: {$exists: false}, nextRunAt: {$lte: this._nextScanAt}, disabled: { $ne: true }},
-          {name: jobName, lockedAt: {$lte: lockDeadline}, nextRunAt: {$lte: this._nextScanAt}, disabled: { $ne: true }}
+          {name: jobName, lockedAt: {$lte: lockDeadline}, disabled: { $ne: true }}
         ]
       },
       {'priority': -1},  // sort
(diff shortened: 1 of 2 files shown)


krishna-koushik:
just wanted to know if this can be merged soon.

PR #316, opened 2016-04-23
feature(agenda): Provide a way to define jobs with regular expressions.

### LINKED ISSUE #209: Distinct, repeatable(every) job instances of the same task method
I need agenda to run the same task at separate times each day, with different data being passed to the task at each run. `agenda.every` is almost perfect for that, except that the same task cannot be repeated, as `every` sets the task type as `single`

To get around that limitation, I started dynamically defining tasks, as and when they are added

```
function addTask (params) {
    agenda.define(params.name, params.config, function (job, done) {
        baseTasks[params.baseTask](job.attrs.data, done);
    });
};
```

This worked fine until an exception caused the server to restart and all the task definitions were lost.

So I tried defining all tasks when the server starts, basically by iterating over jobs in the database, and defining tasks for them. This isn't ideal and caused a lot of bugs.

Finally I just started defining tasks myself, simply copying what `agenda.every` does, omitting setting the type to single.

Would be nice to have this by default.

JonnyBGod:
closes #209

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -79,7 +79,7 @@ Agenda.prototype.database = function(url, collection, options, cb) {
     if (error) {
       if (cb) {
         cb(error, null);
-      } else {
+      } else { 
         throw error;
       }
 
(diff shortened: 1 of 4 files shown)


loris:
Are you sure you made your changes on the master? the diff shows way more changes than just the regex feature

lib/agenda.js
```diff
@@ -556,18 +555,7 @@ function processJobs(extraJob) {
...
-      nextRunAt: job.attrs.nextRunAt,
-      disabled: { $ne: true }
-    };
-
-    var sort = {};
-    var update = { $set: { lockedAt: now } };
-    var options = { new: true };
-
-    self._collection.findAndModify(criteria, sort, update, options, function(err, resp) {
+    self._collection.findAndModify({ _id: job.attrs._id, lockedAt: null, disabled: { $ne: true } }, {}, { $set: { lockedAt: now } }, { new: true }, function(err, resp) {
```
loris: Looks like you reverted most of the commits from past weeks in this PR (ex: https://github.com/rschmukler/agenda/commit/d13ecfdbf630746a39c95dbb1287e16b4f02a2ad)

JonnyBGod:
There are several changes needed for this to work.

1- findAndModify does not work for multiple entries and this features works with multiple jobs
2- change the way to track definitions for the job

I will make a new PR once I figure out why some tests are broken.

PR #317, opened 2016-04-23
feature(agenda): Provide a way to define jobs with regular expressions.

### LINKED ISSUE #209: Distinct, repeatable(every) job instances of the same task method

JonnyBGod:
closes #209

Ended up doing a few tweaks:
- Implemented regular expressions support
- improved _findAndLockNextJob find query (should be more performant)
- updated to use findOneAndUpdate instead of deprecated findAndModify
- commented out `agenda job lock runs a one-time job after its lock expires` test

Regarding the test, I do not understand this particular test. The implementation seams very contradictory. It starts a one time job and expects it to run twice... Am I missing something?

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -437,11 +437,10 @@ Agenda.prototype._findAndLockNextJob = function(jobName, definition, cb) {
   } else {
     this._collection.find({
       name: {$regex: jobName},
-      nextRunAt: {$lte: this._nextScanAt},
       disabled: false,
       $or: [
-        {lockedAt: null},
-        {lockedAt: {$exists: false}},
+        {lockedAt: null, nextRunAt: {$lte: this._nextScanAt}},
+        {lockedAt: {$exists: false}, nextRunAt: {$lte: this._nextScanAt}},
         {lockedAt: {$lte: lockDeadline}}
       ]
     })
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -8,9 +8,6 @@
  *  - Removed db code from jobs.js
  *  - Comments.
  *
- *  TODO:
- *  - Refactor remaining deprecated MongoDB Native Driver methods. findAndModify()
- *
  *  Last change: NF 4/06/2015 2:06:12 PM
  */
 
@@ -37,6 +34,7 @@ var Agenda = module.exports = function(config, cb) {
   this._definitions = {};
   this._runningJobs = [];
   this._lockedJobs = [];
+  this._dbLockedJobs = [];
   this._jobQueue = [];
   this._defaultLockLifetime = config.defaultLockLifetime || 10 * 60 * 1000; //10 minute default lockLifetime
 
@@ -176,7 +174,9 @@ Agenda.prototype.jobs = function( query, cb ){
 
 
 Agenda.prototype.purge = function(cb) {
-  var definedNames = Object.keys(this._definitions);
+  var definedNames = Object.keys(this._definitions).map(function (item) {
+    return '/'+ item +'/';
+  });
   this.cancel( {name: {$not: {$in: definedNames}}}, cb );   // NF refactored 21/04/2015
 };
 
@@ -316,6 +316,7 @@ Agenda.prototype.saveJob = function(job, cb) {
   var id = job.attrs._id;
   var unique = job.attrs.unique;
   var uniqueOpts = job.attrs.uniqueOpts;
+  var disabled = job.attrs.disabled || false;
 
   delete props._id;
   delete props.unique;
@@ -328,7 +329,7 @@ Agenda.prototype.saveJob = function(job, cb) {
       update = { $set: props };
 
   if (id) {
-    this._collection.findAndModify({_id: id}, {}, update, {new: true}, processDbResult );
+    this._collection.findOneAndUpdate({_id: id}, update, {returnOriginal: false}, processDbResult );
   } else if (props.type == 'single') {
     if (props.nextRunAt && props.nextRunAt <= now) {
       protect.nextRunAt = props.nextRunAt;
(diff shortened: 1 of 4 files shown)


jakeorr:
Hi JonnyBGod, that test starts a job that doesn't complete and is picked up again after its lock expires. It simulates a job hanging while in a running state and makes sure that agenda will restart a running job whose lock has expired. In this case it is not contradictory for the job to be run twice.

lib/agenda.js
```diff
@@ -328,7 +329,7 @@ Agenda.prototype.saveJob = function(job, cb) {
       update = { $set: props };
 
   if (id) {
-    this._collection.findAndModify({_id: id}, {}, update, {new: true}, processDbResult );
+    this._collection.findOneAndUpdate({_id: id}, update, {returnOriginal: false}, processDbResult );
```
jakeorr: I can't find any reference to a `returnOriginal` option for `findOneAndUpdate`. Did you mean to use `returnNewDocument`?

lib/agenda.js
```diff
@@ -404,32 +410,71 @@ Agenda.prototype.stop = function(cb) {
...
+    defLimit = defLimit < 0 ? 0 : defLimit;
+  }
+
+  var limit = Math.min(globalLimit || 9999, defLimit || 9999);
+
+  if (limit === 0) {
+    return cb(null, null);
+  } else if (limit === undefined) {
+    limit = 0;
+  }
```
jakeorr: Why is another level of lock limiting happening here? Seems redundant as it is happening [here already](https://github.com/rschmukler/agenda/blob/f25deb84acf843c7ed706aa28fc3ac9c7b4b8f5c/lib/agenda.js#L586) immediately before this method is being called.

lib/agenda.js
```diff
@@ -404,32 +410,71 @@ Agenda.prototype.stop = function(cb) {
...
+        {lockedAt: {$lte: lockDeadline}}
+      ]
+    })
+    .sort({priority: -1})
+    .limit(limit)
+    .toArray(function (err, result) {
+      var jobs;
+      if (!err && result.length) {
+        // Hack: lock pending items
+        self._dbLockedJobs = result.map(function (item ) { return item._id.toString(); });
```
jakeorr: Adding hacks like this suggests to me that there may be a better way to accomplish this.

lib/agenda.js
JonnyBGod: http://mongodb.github.io/node-mongodb-native/2.0/api/Collection.html#findOneAndUpdate

test/agenda.js
```diff
@@ -1012,7 +1012,7 @@ describe("agenda", function() {
         jobs.start();
       });
 
-      it('runs a one-time job after its lock expires', function (done) {
+      /*it('runs a one-time job after its lock expires', function (done) {
```
jakeorr: This test should not be commented.

JonnyBGod: Working on this. Trying to figure out why this test is not passing.

lib/agenda.js
jakeorr: Ah, right. I was looking at the mongo shell docs. Weird they would use the opposite option name like that.  ¯\_(ツ)_/¯

lib/agenda.js
JonnyBGod: This is to due to the lack of an option to "find and modify many". Because I need to do two different queries, one for finding and another for updating, it introduces a lag difference between database and agenda state. This lag was causing a lot of unwanted job runs from "lockOnTheFly". This was the best way to do it. Aldo it s an hack, so I comment it to inform.

lib/agenda.js
JonnyBGod: Because now you can get multiple jobs from this query, we have to limit the query results or you will get more jobs than your limits should permit.

lib/agenda.js
jakeorr: Now that I look closer at what you are doing, it appears that a job could be locked by two different instances.

JonnyBGod: This is just a quick temporary lock to say "Do not touch these jobs while I am still updating them". Worst case scenario, a job will be pushed to the next cycle because of this lag. But this way I prevent a lot of problematic behaviours like duplicated jobs.

JonnyBGod:
Ok understood the test now. Fixed query and it is working now.

lib/agenda.js
JonnyBGod: Ok I see what you mean. Yes this could be a problem. If running multiple instances processing the same job it can result in duplicated jobs. Not sure how to solve this. Will have to think a bit on this.

LucianoGanga:
Hi! 

What's the state of this? Is this option gonna be added soon? 
Thanks!

DVassilev:
@jakeorr @loris Will this be merged or does it not work correctly? Something to implement #209 would be really great

lushc:
@agenda/maintainers Can we get consensus on this PR? Although a lot of good work has gone into this, unfortunately I'm in favour of rejecting it due to the following points:

1. From previous code review these changes seem to break the transactional guarantees when locking a job; because Mongo doesn't have a `findManyAndUpdate` there are two separate queries. So as it stands it's possible for multiple Agenda instances to process the same job (as an aside this is also going to be a potential issue with DB adapters).
2. The issue this PR addresses can be resolved in an easier way.

#209 requires the ability to create multiple jobs with different data using the same job definition name, which people can't do with `every()` because it enforces the job type "single", so only one job with the definition's name can be created. We could either make that behaviour togglable through an option:

```js
agenda.every('15 minutes', 'printAnalyticsReport', {userCount: 100}, {single: false});
```

or introduce a breaking change by renaming the existing method, refactoring accordingly:

```js
// current behavour, creates a single job
agenda.onceEvery('15 minutes', 'printAnalyticsReport');

// multiple "dynamic" jobs
agenda.every('15 minutes', 'printAnalyticsReport', {userCount: 100});
agenda.every('30 minutes', 'printAnalyticsReport', {admin: true});
```

While I prefer the latter option because it's more distinctive, so there's less chance of people overlooking the fact they can create multiple jobs, it's probably better that we preserve BC as this library is getting 1k+ downloads a day. This alongside an update to the docs should hopefully cut down the amount of issues related to dynamic job definitions etc.

Thoughts?

simison:
@lushc feel free to close this, sounds like you have good insight.

@JonnyBGod thanks a lot for helping out with Agenda! Hope you understand that since this is a very popular package, we need to be very careful with API changes. Hope this won't put you off helping out in future.  :beers: 😄

PR #333, opened 2016-06-03
Fix bug in README example snippet

shaialon:
The example without the fix will throw an exception:
*SyntaxError: Unexpected token )*


loris:
Thanks!

shaialon:
sure

Issue #296, opened 2016-03-08
working with latest version of aganed and mongo 2.4


nbukai:
Hello,

I have just started to work with agenda. The environment I'm working in include mongo version 2.4 which I cannot upgrade. Is there any way to work with latest agenda version and mongo 2.4? 

Thanks for the support.

fsamir:
Check this out:
https://github.com/rschmukler/agenda/pull/334

rschmukler:
Closed by #353

Issue #330, opened 2016-05-22
Bump mongodb version?


coodoo:
`agenda` is currently using a slightly older version of mongodb which doesn't support async `toArray()` with `find()`, upgrading to the latest version will fix it, could you please bump it? Thanks.

fsamir:
Check this out:
https://github.com/rschmukler/agenda/pull/334

rschmukler:
Closed by #353

Issue #299, opened 2016-03-11
mongo driver 2.1 support for Mongodb3.2


mindfieldx:
Could be possible upgrade mongodb library to 2.1.x to support Mongodb3.2 ?
Thanks

toddbluhm:
We were having issues with the mongodb driver 2.1.8 in production (I believe agenda uses 2.0.34, but we wanted to be on the same driver across all dependencies) so I forked agenda and bumped it up to 2.1.10 and we seem to be having no issues so far. I would suggest you do something similar until agenda gets updated with the most recent version.

ianvonholt:
As with @toddbluhm, I forked and am running using mongodb at 2.1.15 without problems.

fsamir:
Check this out:
https://github.com/rschmukler/agenda/pull/334

joeframbach:
Related issue: https://github.com/joeframbach/agendash/pull/29/files

rschmukler:
Closed by #353

PR #334, opened 2016-06-07
Update to mongo driver 2.1.11 and extend mongo compatability to inclu…

philbrookes:
## What
- Update agenda to use 2.1.11 of the node.js mongo driver.
- Update agenda to be compatible back as far as mongo 2.4.

## Why
- 2.1.11 of the Mongo driver is the most recent release and compatible upto the latest releases of mongo.
- Mongo 2.4 is still quite widely used and adding support for this version of mongo increases the use-cases for agenda.

## How 
- The upgrade of the mongo driver was simply updating the package.json file, and retesting the package
- Compatability with 2.4 was broken by the call to createIndexes, catching the exception here, and gracefully failing back to ensureIndex which is compatible with 2.4 resolved the issue. 

## Testing
This has been tested against Mongo 2.4 and 3.2, and also tested with node 0.10.x and 4.4.x. The testing was executed using `npm test` in each scenario.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,3 +1,7 @@
+0.9.0 / 2016-06-07
+==================
+  * Upgrade mongo driver to 2.1.11
+  * Increase Mongo compatability to 2.4
 
 0.8.1 / 2016-05-08
 ==================
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -106,14 +106,32 @@ Agenda.prototype.db_init = function( collection, cb ){
                                   "key": {"name" : 1, "lockedAt" : 1, "priority" : -1, "nextRunAt" : 1, "disabled" : 1},
                                   "name": "findAndLockNextJobIndex2"
                                 }],
-                                function( err, result ){
-                                  if (err) {
-                                    self.emit('error', err);
-                                  } else {
-                                    self.emit('ready');
-                                  }
+                                function (err, result) {
+                                  try {
+                                    if (err) {
+                                      self.emit('error', err);
+                                    } else {
+                                      self.emit('ready');
+                                    }
+                                    if (cb) cb(err, self._collection);
+                                  } catch(e){
+                                    if(e.message === 'no such cmd: createIndexes'){
+                                      // Looks like a 2.4.x database
+                                      self._collection.ensureIndex(
+                                        {"name": 1, "priority": -1, "lockedAt": 1, "nextRunAt": 1, "disabled": 1},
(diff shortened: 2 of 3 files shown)


grdryn:
This potentially fixes #296, #299, and #330.

david-martin:
@rschmukler This looks like a good fix to get in & published

rschmukler:
Sorry on the delay (and thank you @david-martin for the poke). This looks good to me. Final call for @loris, @jakeorr or anybody else to offer commentary before I merge it in...

demianhh:
Thanksgiving

philbrookes:
Hi @rschmukler, can this PR be merged now?

Cheers, Phil.

fsamir:
Hi guys,
I would like to point out that the code in this PR doesn't work properly with mongo 2.4. because 'ready' event is never fired from within the catch clause. `self.emit('ready');`

Anyone using mongo 2.4 and waiting on the the `ready` event to initialise agenda, will face this issue:

````
agenda.on('ready', function () {
   ...
   agenda.start();
   ...
````

Also, I would suggest you drive the logic based on the version of mongo, not on an exception. In addition, it would be important to not fire 'error' (it currently always fire it on mongo 2.4) until the fallback code has also failed.

**Comment by fsamir** [comment]
I've made the afore mentioned fixes and raised another PR:
https://github.com/rschmukler/agenda/pull/340

If you need to pull it before it is merged in the original project, just do this in package.json:
```
"agenda": "https://github.com/fsamir/agenda.git#2d1832a8a0bbb1180f134da1d3537c8a1c00b9c6"
```

loris:
IMO, indexes should not be created automatically by the library. It should be something recommended as part of the setup in the README.md

simison:
@philbrookes perhaps you could have a look at @fsamir PR #340 and maybe even include those fixes at this PR?

malas:
why is this still not merged? @rschmukler 

For all the users of Mongo in ARM devices, the only official version is 2.4. And this PR does not brake anything at all. Please make this merge.

rschmukler:
@malas (and everyone) - I haven't merged this yet because I agree with @loris' comments on indexes not being auto-created, while this PR introduces the bug mentioned (and fixed by) @fsamir. If somebody would submit a PR with these changes and the bug fixed by @fsmair for Mongo 2.4, I'd be happy to merge it. I haven't had time to implement it myself yet.

malas:
@rschmukler i added #353 pull request as you asked. The build passes.

PR #355, opened 2016-08-19
Fix for race condition in the afterEach clean up code

C45tr0:
The race condition would happen after test "job concurrency" -> "should not block a job for concurrency of another job"
The change made it stop processing jobs before disconnecting.


rschmukler:
Thanks!

PR #361, opened 2016-09-08
Run job between start and end date + run job every x days

suryanaik:
Modified job.repeatEvery to include startDate, endDate and skipDays in the option to achieve the following scenarios:
- Every 45 days at 8:00 AM CST starting 30-Aug-2016, run till 30-Aug-2017
- Daily at 8:00 AM CST starting 30-Aug-2016, run till 30-Aug-2017
- Every Wed and Fri at 8:00 AM CST Starting 30-Aug-2016, run till 30-Aug-2017

Will not disrupt existing users.


simison:
Hmm, you've changed the indentation from 2 to 4 all over the place.

suryanaik:
Ah! I didn't realize that. Sorry about that.

wtfuii:
I think you need to fix that indentation issue before your PR will be reviewed and merged. I need that feature too, so i would be glad if the merge happens soon.

suryanaik:
I have fixed the indentation for job.js but not able to fix it for the test/agenda.js. Can some one help?

OmgImAlexis:
@suryanaik if you'd like to rebase this I'll be happy to review and look at getting this merged.

leonardlin:
will try to update this PR to work with current TS version.

lib/agenda/index.ts
```diff
@@ -101,7 +101,7 @@ class Agenda extends EventEmitter {
     super();
 
     this._name = config.name;
-    this._processEvery = (humanInterval(config.processEvery) ?? humanInterval('5 seconds')) as number;
+    this._processEvery = (humanInterval(config.processEvery) ?? humanInterval('5 seconds')) as number; // eslint-disable-line @typescript-eslint/non-nullable-type-assertion-style
```
leonardlin, line 104: This must be a new lint-rule that came with the latest dep-update.

Unfortunately I don't understand the rule. I also don't think it applies to this line.

lib/job/compute-next-run-at.ts
```diff
@@ -34,40 +34,68 @@ export const computeNextRunAt = function(this: Job) {
...
     if (timezone) {
       cronOptions.tz = timezone;
     }
 
     try {
       let cronTime = parser.parseExpression(interval, cronOptions);
-      let nextDate = cronTime.next().toDate();
-      if (nextDate.valueOf() === lastRun.valueOf() || nextDate.valueOf() <= previousNextRunAt.valueOf()) {
+      let nextDate: Date | null = cronTime.next().toDate();
+      if (nextDate.getTime() === lastRun.getTime() || nextDate.getTime() <= previousNextRunAt.getTime()) {
```
leonardlin, line 49: Consistent type.
Variables used to be mixed between: number | Date | moment.Moment

lib/job/repeat-every.ts
```diff
@@ -7,9 +7,14 @@ import { Job } from '.';
  * @param interval repeat every X
  * @param options options to use for job
  */
-export const repeatEvery = function(this: Job, interval: string | number, options: any = {}): Job {
+export const repeatEvery = function(this: Job, interval: string | number, options: { timezone?: string; startDate?: Date | number; endDate?: Date | number; skipDays?: string; skipImmediate?: boolean } = {}): Job {
```
leonardlin, line 10: added type information for the options object

test/agenda.js
```diff
@@ -264,15 +264,15 @@ describe('Agenda', function() { // eslint-disable-line prefer-arrow-callback
           await jobs.every('5 minutes', jobName, {}, { skipImmediate: true });
           const job = (await jobs.jobs({ name: jobName }))[0];
           const nextRunAt = job.attrs.nextRunAt.getTime();
-          const now = new Date().getTime();
+          const now = Date.now();
```
leonardlin, line 267: new linting rule complained about it.

lib/agenda/index.ts
simison: I had a quick look and didn't understand it either.

Added lint-ignore here too https://github.com/agenda/agenda/pull/1182/files#r549456554

lib/job/compute-next-run-at.ts
```diff
@@ -34,40 +34,68 @@ export const computeNextRunAt = function(this: Job) {
...
-          this.attrs.nextRunAt = (lastRun as moment.Moment).valueOf() + (humanInterval(interval) ?? 0);
-          debug('[%s:%s] nextRunAt set to [%s]', this.attrs.name, this.attrs._id, new Date(this.attrs.nextRunAt).toISOString());
+          this.attrs.nextRunAt = new Date(lastRun.getTime() + (humanInterval(interval) ?? 0));
+          debug('[%s:%s] nextRunAt set to [%s]', this.attrs.name, this.attrs._id, this.attrs.nextRunAt.toISOString());
         }
       // Either `xo` linter or Node.js 8 stumble on this line if it isn't just ignored
       } catch {}
     } finally {
-      if (Number.isNaN(this.attrs.nextRunAt)) {
+      if (Number.isNaN(this.attrs.nextRunAt.getTime())) {
```
simllll, line 98: i think this check was for checking if "date" is valid. now if nextRunAt is not set (which can be the case as far as I remeber) it could break.. maybe we can fix this in an upcoming PR when you added the types for attrs (and make nextRunAt optional, typescript should complain)

lib/job/compute-next-run-at.ts
```diff
@@ -80,15 +108,15 @@ export const computeNextRunAt = function(this: Job) {
    */
   const computeFromRepeatAt = () => {
     const lastRun = this.attrs.lastRunAt || new Date();
-    const nextDate = date(repeatAt).valueOf();
+    const nextDate: Date = date(repeatAt);
```
simllll, line 111: before casting "any/unknown" to "Date" I would recommend adding a return type for "date" (not quite sure where date is coming from, if it is from an external library, you could add a wrapper for this function or use a type guard.

lib/job/repeat-every.ts
```diff
@@ -7,9 +7,14 @@ import { Job } from '.';
...
   this.attrs.repeatInterval = interval;
   this.attrs.repeatTimezone = options.timezone ? options.timezone : null;
+  // Following options are added to handle start day
+  // and cases like run job every x days (skip some days)
+  this.attrs.startDate = options.startDate ?? null;
```
simllll, line 15: maybe it's intended, just wanted to point it out: startDate can either be Date | number or undefined..and here undefined is "converted" to null. I usually try to keep things either undefined or null, but not both for "same" purposes of having things "not defined". Or was here any specific reason to set startDate to "null" by default then? (Ohterwise I think just remove ?? null would be fine too)

simllll:
overall looks good to me, just two comments, but nothing that is blocking at all :+1:

lib/job/compute-next-run-at.ts
leonardlin: @simllll  thanks. The 'date' is valid check is still needed here. There are 3-4 test-cases testing this flow.
In the old code `this.attrs.nextRunAt` was `this.attrs.nextRunAt.getTime()`.
`this.attrs.nextRunAt` is "Invalid Date".
To test if a date is invalid, you have to use `isNan(date.getTime())`

https://stackoverflow.com/questions/1353684/detecting-an-invalid-date-date-instance-in-javascript

lib/job/repeat-every.ts
leonardlin: I don't have a preference for null vs undefined.

I reuse the code from the original PR.
I kept the existing style.
for example `timezone` is checked for undefined and set to null.
I'm fine either way.

lib/job/compute-next-run-at.ts
leonardlin: yeah, good point.
I will address this when I add more type information.

leonardlin (approved):
will continue to add more types in separate PR

lib/job/compute-next-run-at.ts
simllll: I see your point, i just thought that nextrunat could be undefined already,which will let the code fail if you access a property/method on "undefined". But when it was the case before, its probably good to go (y)

PR #368, opened 2016-09-22
Bump mongodb dep version to support ssl conns

adrukh:
As described in https://github.com/rschmukler/agenda/issues/359, using a 2.1.x version does not allow ssl connections to mongodb.

This PR bumps the dependency to a version that supports ssl. It also follows semver `^` to allow further dependency version upgrades without modifying `package.json`.

Test are passing locally with node 6.6.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "date.js": "~0.3.1",
     "human-interval": "~0.1.3",
     "moment-timezone": "^0.5.0",
-    "mongodb": "2.1.11"
+    "mongodb": "^2.2.10"
   },
   "devDependencies": {
     "blanket": "1.1.5",


ragulka:
This should probably also help with the dreaded `TypeError: Argument must be a string` when using different mongodb versions, and doing something like:

```javascript
var ObjectId = require('mongo').ObjectId;
agenda.jobs({ ObjectId( jobId ) }) //...  throws TypeError above

// temp fix:
var ObjectId = require('agenda/node_modules/mongodb').ObjectId;
```

cojack:
@rschmukler halo?

adrukh:
Thanks for merging! Any ETA on publishing a new version on npmjs.com?

PR #379, opened 2016-11-06
Fixes + protects against concurrency not being honored 

netslayer:
We had an issue where the job done callback was being over executed. This made the global job level running and lock counters to go negative. Then agenda would stop obeying concurrency limits. This prevents the job counter from ever going negative and adds an exception if code calls the job done more than once.


rschmukler:
Thank you!

PR #383, opened 2016-11-18
Return failure report not just reason message

gastonelhordoy:


simison:
@lushc commented #441:
>I'm not sure of. I haven't come across this use-case myself yet but I'd imagine if you needed more context, couldn't reason just be an object you've passed in with all the attributes you need, since it can be stored as-is in MongoDB? Low priority.

@gastonelhordoy sorry I know this is pretty aged PR, but wanna elaborate a bit your reasoning for this?

@lushc feel free to close if you think this isn't needed.

leonardlin:
This PR has been implemented in the latest version.
PR not necessary anymore.

PR #408, opened 2017-03-01
Added .isJob method

Firanolfind:
Method for to detect Job instance.
https://github.com/rschmukler/agenda/issues/407

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -137,6 +137,10 @@ Agenda.prototype.name = function(name) {
   return this;
 };
 
+Agenda.prototype.isJob = function(obj) {
+  return obj instanceof Job;
+};
+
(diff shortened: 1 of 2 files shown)


simison:
@lushc commented at https://github.com/agenda/agenda/issues/441#issuecomment-309106377:
> It doesn't seem to have much of a use-case outside of the contributor's initial issue of using jobs as part of other models, and then not being able to determine if the job object was created by Agenda or not.

I second that.

This would need more +1's to go forward.

simison:
Should we _not_ implement this, @agenda/maintainers ?

I don't see common use case.

OmgImAlexis:
Since it's just a wrapper for `instanceof` I see no value in this.

Issue #422, opened 2017-04-20
Scheduling job with agenda.every method with human interval runs the job immediately


jagadish-kb:
Scheduling the job with agenda.every method with human interval runs the job immediately and not after at the defined interval. 
For ex., If i would like to run a Job every '4 hours', with the option 
```
agenda.every('4 hours', 'MY_JOB_NAME');
```
I see that the 'MY_JOB_NAME' runs immediately. Not sure why it is required to run the job immediately when scheduling via human interval. Even if it is required, can we atleast have an option where we can skip this immediate run? 

I did a minor fix in the function 'computeFromInterval' in job.js and it skips running this immediately. PFB fix:

```
function computeFromInterval() {
    var lastRun = this.attrs.lastRunAt || new Date();
    .....
      // Nope, humanInterval then!
      try {
        //Just removed the special check if nextRunAt is !present 
         this.attrs.nextRunAt = lastRun.valueOf() + humanInterval(interval);
      } catch (e) {}
    ....
  }
```
If at all, the immediate run is also required then `computeNextRunAt` method could be passed the `options` attrib which could have a flag `skipImmediate=true` which could be put in the above block of fix and both behaviors can co-exist. But I did see few other defects where people didnt want this default run to happen immediately. So kindly advise.

Neoflash1979:
@jagadish-kb Not only with human interval, but also with numbers. Unless I did something wrong, I'm having a problem where the job will run immediately and only once. It won't run again at the selected interval. Here is my code:

```
const dburl = 'mongodb://localhost:27017/nightlife';
const url = process.env.MONGOLAB_URI;

import * as Agenda from 'agenda';

const agenda = new Agenda({
  db: { address: url || dburl, collection: 'agendaJobs' },
  processEvery: '60 seconds'
});

agenda.define('print to console', function (job, done) {
  console.log('Job just fired');
});



agenda.on('ready', function () {
  agenda.every(90000, 'print to console');
  agenda.start();
});

function graceful() {
  agenda.stop(function () {
    console.log('Exiting Agenda')
    process.exit(0);
  });
}

process.on('SIGTERM', graceful);
process.on('SIGINT', graceful);
```

lushc:
It looks like the behaviour for running a human interval job immediately was introduced in dd5bd407.

arihantdaga:
In the new library version 1.0.3 its is still happening..

marcusjwhelan:
[...]

PR #423, opened 2017-04-20
repeatInterval with human-interval was running the jobs immediately

jagadish-kb:
Fix for issue https://github.com/rschmukler/agenda/issues/422 . Made this behavior (on/off) configurable and existing behavior remains as default behavior & whoever wants this fix must explicitly set the option `skipImmediateRun` to true.


nemozny:
Yes! Thank you @jagadish-kb , that would save me some considerable troubles as I would not need to manage this behavior on application level.

leonardlin:
this was solved with version 2.0
https://github.com/agenda/agenda/issues/422

Issue #441, opened 2017-06-16
Triaging issues and pulls


simison:
It's a bit much for one person to go trough everything.

Any ideas/experiences how this should be done? Using tags or just list here (feel free to modify this post)?

Before anything is merged, should we have at least one "LGTM" from anyone in the team?

Everything should probably be sorted into:
- high priority
- low priority
- should be closed

lushc:
Here are my opinions based on me poking around the repo the past few days and incorporating some of these PRs:

* #268 can be rebased and merged. There's a small conflict where the `lockOnTheFly()` query was refactored, [see this commit for reference](https://github.com/Codevate/agenda/commit/048c7f14e89e5fdcfe86dc74cc7aae42d7948bfe#diff-1ba7680cb21f30bb02af3c7899c41248L582). High priority.
* #241 needs more investigation. Although it [fixed the recursive lock issue for sshen81](https://github.com/agenda/agenda/issues/104#issuecomment-161750668) I still got errors when using the [benchmarking script](https://github.com/nwkeeley/agenda/blob/56013e472df6f08fa9ed0902aa2706f55dddd9bd/script.js) mentioned in #104, but maybe that's because the index order wasn't adjusted. That said, it also partially reverts the reasoning for the query changes in #180 and could be made redundant by the next PR. Low priority.
* #300 needs more work as I noticed [two problems after rebasing](https://github.com/agenda/agenda/pull/300#issuecomment-308211215) and it might benefit from a [query and index order tweak](https://github.com/Codevate/agenda/commit/d89b58c7cd43a78e39accb534eccb20c48fbb6b8). It completely solved #104 for me and potentially closes similar performance issues such as #437. High priority.
* #423 might be closed in favour of better documentation. The main issue is that using `repeatEvery()` in isolation doesn't set the `nextRunAt` time, but calling`computeNextRunAt()` doesn't work either since `lastRunAt` has yet to be set for new jobs, so the interval is ignored. [This code example](https://github.com/agenda/agenda/issues/192#issuecomment-140126659) could be added to the documentation as an example of delaying the execution of a repeat job. Low priority.
[...]

Issue #414, opened 2017-03-22
Maintainers needed?


simison:
Heyy @rschmukler,

I really like Agenda and I use it a lot! :beers: Just pity that maintaining this package isn't very active anymore. :-(

There seems to be enough interest from community to send pull requests and open issues tho, so perhaps we could find some folks to help with looking after PRs and issues?

WDYT?

Unfortunately I'm already super busy with other open source stuff myself.

Anyone else interested?

gastonelhordoy:
I'll be more than willing to help out and contribute if needed. I'm also heavily relying on this project so it would be good to get some progress on PRs and issues.

Let us know your thoughts on this @rschmukler

Firanolfind:
Agree. +1

rschmukler:
Hey all! I would be ecstatic to get some help maintaining. I deeply apologize for the lack of maintenance. I recently founded a company and have largely transitioned away from Node so my efforts have been limited as late. 

How would you all suggest I proceed from here? Just grant access to a few people on the repo?

PS I will publish #395 tonight

rschmukler:
Hey all. I just got Node re-set up on my home computer. This means that I should be better equipped to handle maintenance issues around agenda.

That being said, I would still love help maintaining (and even further developing) this package!

If anyone wants to get involved, please feel free to post here or email me at `ryan@slingingcode.com`. I propose the following: People who want to maintain, please contact me and also start replying to issues / PRs with a LGTM etc. Like I said, I won't being reviewing too much but once we get a consensus going I will keep note of the interested parties who are helping along. From there, I (or you, the community) can pick a few members to get NPM publishing / GitHub write access so that maintenance can continue along at a more rapid speed and is no longer dependent on me as a gate keeper.

PS. Existing maintainers will be kept on (@loris, @nkweekly and @droppedoncaprica) in case they also want to help contribute.

I'm very excited to get others involved in this, so please don't hesitate to join in!

bonesoul:
hope this package can get the love from community that it deserves

droppedoncaprica:
I totally agree that this package needs a more active maintainers to keep this running.  I've moved to a new job that uses a completely different back-end myself, so I've not been keeping up with this library's issues / pull requests as much as I have in the past. 

If anyone needs help with maintenance, feel free to @ ping me and I'll do my best to take a look at it!

OmgImAlexis:
@rschmukler maybe look at moving this to an org so it's easier to handle others getting merge rights, etc?

[...]

Issue #402, opened 2017-02-02
repeatAt setting is not respected


derickbailey:
I've got a job set up with the following:

```js
var loadInterval = "1 day";
var loadTime = "15:00";

var job = agenda.create("schedule.load");
job.schedule("next " + loadTime)
job.repeatEvery(loadInterval, {
      timezone: TIME_ZONE
});

job.repeatAt(loadTime);

job.save(cb);
```

when I set my clock to 3pm in the correct timezone, the job runs. upon completion - a few seconds later - the job is rescheduled and set to run the next day ... at 3pm + a few seconds.

if I schedule this job for 3pm, and then set my clock to 5:15pm in the correct timezone, the job will run as expected, because the time has passed and it needs to run.

when the job completes, however, it sets the next job run for 1 day later, at 5:15pm + a few seconds.

the `repeatAt` time is not being respected. 

i expect the job to repeat every day, at 3pm, no matter when it actually ran. that's why i'm calling `repeatAt`... to ensure it repeats at that specific time. but it is repeating 1 day after running, no matter when the actual run time occurrs, instead of respecting the `repeatAt` setting.

is this expected?

if so, how do i achieve my desired result: every day at 3pm (no matter when the actual job runs - always 3pm, next)?

Conroman16:
Bump?  @rschmukler Any word on why this is the case?  I'm having an issue with `repeatAt` as well

Yooo84:
@rschmukler Have you found a solution because I have the same problem?
Thx

Conroman16:
@Yooo84 

Looks like @rschmukler is pretty busy these days and is unable to provide this repo the level of support it needs.

As for a solution to the problem, I don't have one that's worth anything.  I actually ended up just working around the issue by scheduling the jobs as singletons and then writing some extra code to manually reschedule them as new singletons....  lather, rinse, repeat

It's certainly not a clean solution, however since @rschmukler is really busy these days and has stated he's open to contributions... I'm going to take a look and potentially submit a PR to fix it as I'd love to get rid of my workaround code

simison:
@Conroman16 see #414 and follow ups #441 if you'd like to help!

lushc:
I believe this is due to both `repeatEvery()` and `repeatAt()` being called. Because an interval is set, when the job computes its next run time it'll be using that rather than the repeat at time:

```js
  if (interval) {
    computeFromInterval.call(this);
  } else if (repeatAt) {
[...]

PR #443, opened 2017-06-20
Db adapters

### LINKED ISSUE #18: Database Adapters
Can you allow extensibility to this package to hook up different data stores, so that we can add additional database adapters to it. I am looking for MYSQL support and if it was allowed, then we could use this package for different data sources.

C45tr0:
Closes #18 

I have been testing this with DynamoDB for awhile now. While my DynamoDB adapter has a lot of issues, I think this is at a decent point as a default setup. 

This should be none breaking as the mongo db adapter is still in the repo. 

Let me know your thoughts.

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -45,7 +45,11 @@ var Agenda = module.exports = function(config, cb) {
   if (!config.adapter || config.adapter == 'mongodb') {
     DBAdapter = require('./db_adapters/mongodb');
   } else {
-    DBAdapter = require("agenda-db-" + config.adapter);
+    if (typeof config.adapter === 'string') {
+      DBAdapter = require("agenda-db-" + config.adapter);
+    } else {
+      DBAdapter = config.adapter;
+    }
   }
 
   this._dbAdapter = new DBAdapter(this, config, cb);
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -374,9 +374,8 @@ Agenda.prototype._findAndLockNextJob = function(jobName, definition, cb) {
 
   // Don't try and access Db if we've lost connection to it. Also see clibu_automation.js db.on.close code. NF 29/04/2015
   // Trying to resolve crash on Dev PC when it resumes from sleep.
-  var s = this._mdb.s || this._mdb.db.s;
-  if (s.topology.connections().length === 0) {
-    cb(new Error( 'No MongoDB Connection'));
+  if (!this._dbAdapter.hasConnection()) {
+    cb(new Error( 'No DB Connection'));
   } else {
     this._dbAdapter._findAndLockNextJob(jobName, this._nextScanAt, lockDeadline, function (err, result) {
       var job;
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -19,9 +19,7 @@ var Job = require('./job.js'),
   utils = require('util'),
   Emitter = require('events').EventEmitter;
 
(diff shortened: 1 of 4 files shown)


simison:
Closing end reopening to run Travis tests (we enabled them again just recently).

simison:
@C45tr0 This one is reeeally interesting. However, I'd be really afraid this adds complexity right now when this package isn't maintained very actively (we're at it, see #414 and #441).

Would you be interested in hopping in and working on Agenda actively in future as well, perhaps bake in that DynamoDB adapter as well?

Also, needs rebasing. ;-)

lushc:
46ff788, 9b975e0, and 228464c (except [lib/agenda.js](https://github.com/agenda/agenda/pull/443/commits/228464c505dcd1fc3de18fcdebeffaaf7dbe6b2c#diff-1ba7680cb21f30bb02af3c7899c41248)) should be a separate PR since these commits might address #402 by adding a useful new feature, and potentially squish similar timezone bugs for `repeatAt` which were fixed for `repeatEvery` in #281.

I can see you've put in a lot of work abstracting out the various DB calls and having the choice of adapter config-driven, commendable effort. Could you give more insight into your DynamoDB implementation and the issues you're currently facing? My main concern is that we might encounter some weird edge-cases, given how Agenda's locking & queuing behaviour (`_findAndLockNextJob` and `lockOnTheFly`) is designed around MongoDB's capabilities. For example:

> findOneAndUpdate() updates the first matching document in the collection that matches the filter. The sort parameter can be used to influence which document is updated.

Because this is a single atomic operation with a write-lock for the duration, it's guaranteed that no other instance of Agenda can pick up the same job and run it, so every other adapter would have to honour the same guarantee in some way. https://github.com/agenda/agenda/issues/18#issuecomment-133440122 mentions that locking behaviour could be different in Redis, for example, which might mean it can't fit into the current design.

I guess my main concern right now is that without another adapter to go with this PR, one that's proven to be as reliable as the existing implementation with some supporting documentation on the internals & adapters to get contributors up-to speed, we might end up merging but not have any other adapter written. I'd be really interested in hearing more about your tests and how close you are to achieving 1:1 behaviour. I've only been using this library a short while but my gut feeling right now is that we're very tightly coupled to MongoDB.

C45tr0:
I agree with the separate pull request for the timezone stuff. I am planning on trying to get something for that this weekend. To be honest, I forgot that I added that myself. I will look and see what I can do.

For DynamoDb, the adapter was a first draft, and I only got it working enough for me to move on. The _findAndLockNextJob is the reason I am working to move away from DynamoDB for my needs. Mainly due to the different situations it needs to query on and how DynamoDB handles reading its data. Maybe someone with a better understanding of DynamoDB could implement it, but for now I don't have any solutions that would work. If I get some time in the coming weeks, I might look back into it.

I have not thought through the atomicness of the operations. I am currently working on a Sequelize version, which will support MySQL 5.7, Postgres 9.5, and SQLite 3. I have already noticed some issues with it and not having a single atomic operation to get and update. Basically not having that, causes race conditions. I will be looking into that over the next week or so.

I mainly wanted to get the pushed to start getting feedback. I figured there would be at least a little bit of push back/requests. I appreciate all of the feedback so far.

@simison I am always willing to help. I will try and send pr's on things I find and fix. Not sure about the DynamoDB stuff, as it needs a lot of work still and I am not sure it will correctly in the end anyway.

PR #445, opened 2017-06-21
Update Agendash links

simison:
Is `agenda-ui` maintained anymore, should it be removed?


joeframbach:
agenda-ui has been updated to state it is deprecated and now points to agendash. The link can be removed.

simison:
@joeframbach done!

PR #268, opened 2016-01-18
replaced the deprecated findAndModify method from native js driver to…

etsalah:
I have tried to migrate that library from using findAndModify method. However the tests usually fail on second execution. if someone can help fix that issue so that the pull request can be accepted. Need this to be able to use it in production environment because we are getting errors because we can seem to assign any role that has findAndModify to db user after implementing access control

PR #331, opened 2016-05-23
Add support for `mongoose.connection` for `agenda.mongo()`, fixes #156

kerimdzhanov:
This PR adds support of `mongoose.connection` as `agenda.mongo()`.

Now you can share your existing mongoose connection with agenda instances, like:

```js
var mongoose = require('mongoose');
var Agenda = require('agenda');

var agenda = new Agenda({ mongo: mongoose.connection });
```

PR #448, opened 2017-06-22
replaced the deprecated findAndModify method from native js driver to…

lushc:
… findOneAndUpdate

Closes #268

### DIFF
--- a/lib/agenda.js
+++ b/lib/agenda.js
@@ -343,7 +343,7 @@ Agenda.prototype.saveJob = function(job, cb) {
       update = { $set: props };
 
   if (id) {
-    this._collection.findAndModify({_id: id}, {}, update, {new: true}, processDbResult );
+    this._collection.findOneAndUpdate({_id: id}, update, {returnOriginal: false}, processDbResult );
   } else if (props.type == 'single') {
     if (props.nextRunAt && props.nextRunAt <= now) {
       protect.nextRunAt = props.nextRunAt;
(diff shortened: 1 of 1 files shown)


simison:
Node.js v0.11 failed tests — how about we just ditch v0.10 and v0.11 going forward?

lushc:
I'm not entirely sure why it failed and if it's down to the Node version, [the native driver still supports v0.10 and v0.12](https://github.com/mongodb/node-mongodb-native/blob/2.2/.travis.yml), so I'm assuming it was just an intermittent connection issue ([they mention this in the docs](https://docs.travis-ci.com/user/database-setup/#MongoDB-does-not-immediately-accept-connections)) which backs up the actual build error.

simison:
@lushc if you wanna debug/test that, you can close+re-open to re-run Travis.

lushc:
Ah, that'd be much quicker than testing it in a TravisCI docker container like I just did...

lushc:
@simison all passing this time, rather than removing these versions do you think we should mimic the native driver and test against 0.12 instead of 0.11? Although they have been end-of-life since late 2016, so maybe removing them is better, not sure.

simison:
0.12 might just pass or bounce again randomly, I'm not sure if it's really worth effort. They're end-of-life for exactly this reason that people wouldn't need to keep supporting them as it's time consuming. ;-)

simison:
...I'd merge this without rebasing and then merge #449 later if no-one has anything against it.

lushc:
Oh I didn't remove 0.11 by the way, so #449 can be merged whenever. The build failure was just an intermittent issue where Mongo wasn't ready by the time the tests started running, which is a bit odd as the build already waits 15 seconds for it. This has happened to me before on other projects so I'm probably cursed :P

simison:
@lushc apparently the same issue with Node.js v0.11 occurred here: https://github.com/agenda/agenda/pull/331#issuecomment-221035610

simison:
Thanks a bunch btw for this! Happy to get some actual code merged in finally. :-)

lushc:
No problem, you've really managed to get the ball rolling on triaging issues/PRs and getting an active core of maintainers together, so thank you 👍 🍺  a remarkable difference compared to early last week!

lib/agenda.js
```diff
@@ -427,17 +427,16 @@ Agenda.prototype._findAndLockNextJob = function(jobName, definition, cb) {
...
         $or: [
           {name: jobName, lockedAt: null, nextRunAt: {$lte: this._nextScanAt}, disabled: { $ne: true }},
           {name: jobName, lockedAt: {$exists: false}, nextRunAt: {$lte: this._nextScanAt}, disabled: { $ne: true }},
           {name: jobName, lockedAt: {$lte: lockDeadline}, disabled: { $ne: true }}
         ]
       },
-      {'priority': -1},  // sort
       {$set: {lockedAt: now}},  // Doc
-      {'new': true},  // options
+      {returnOriginal: false, 'priority': -1},  // options & sort
```
lushc: I've just noticed that according to the [findOneAndUpdate docs](http://mongodb.github.io/node-mongodb-native/2.2/api/Collection.html#findOneAndUpdate) this is incorrect. It should be:

```js
{returnOriginal: false, sort: {priority: -1}}
```

There doesn't seem to be any test coverage for this either. I'll put in another PR to fix.

simison: Follow ups https://github.com/agenda/agenda/pull/451

PR #451, opened 2017-06-27
Fix findOneAndUpdate sort option and add test coverage...

lushc:
…for priority running order.

It looks like the existing indexes meant documents were returned in the correct priority order anyway, but it's better to fix this so the sort order is guaranteed.

### DIFF
--- a/test/agenda.js
+++ b/test/agenda.js
@@ -1182,6 +1182,33 @@ describe("agenda", function() {
         }, 100);
       });
 
+      it('should run the higher priority jobs first', function (done) {
+        var results = [];
+        var now = new Date();
+
+        jobs.define('priority', { concurrency: 1 }, function (job, cb) {
+          results.push(job.attrs.priority);
+          setTimeout(cb, 50);
+        });
+
+        var normal = jobs.create('priority').schedule(now);
+        var low = jobs.create('priority').schedule(now).priority('low');
+        var high = jobs.create('priority').schedule(now).priority('high');
+
+        normal.save(function () {
+          low.save(function () {
+            high.save(function () {
+              jobs.start();
+
+              setTimeout(function () {
+                expect(results).to.eql([10, 0, -10]);
+                done();
+              }, 150);
+            });
+          });
+        });
+      });
+
     });
 
     describe("every running", function() {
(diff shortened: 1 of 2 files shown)


simison:
LGTM

test/agenda.js
```diff
@@ -1182,6 +1182,33 @@ describe("agenda", function() {
...
+        var normal = jobs.create('priority').schedule(now);
+        var low = jobs.create('priority').schedule(now).priority('low');
+        var high = jobs.create('priority').schedule(now).priority('high');
+
+        normal.save(function () {
+          low.save(function () {
+            high.save(function () {
+              jobs.start();
+
+              setTimeout(function () {
```
simison: There might be a better way to get called when jobs are done, e.g. when `agenda.on('complete')` gets called 3 times? Or with `complete:job name` event?

https://github.com/agenda/agenda#job-queue-events

simison:
@lushc this would be ready to be merged? I don't really have time to test this myself but I see you added a test for this case so that's a big +1

lushc:
This will be superseded by #464, but hopefully the test can be cherry-picked over.

test/agenda.js
niftylettuce: Thanks @simison and @lushc I'm implementing this right now using `removeListener` as well (for the count = 3 approach). If it doesn't get a count of 3 after X period of time (whatever is default mocha timeout) then it will fail ofc.  See my PR in #462 for the tests to be added there.  Thank you.

niftylettuce: I think I'm actually going to use `on('start:priority` instead.  If anyone has feedback against that please let me know.  This ensures that the start is in order, as opposed to complete events.

niftylettuce: Oh goodness cc @OmgImAlexis this test file is absolutely horrid JavaScript 😆 hahahah!

niftylettuce: OK fixed all eslint warnings in test file, now wrapping up tests for #462

niftylettuce:
I added tests here https://github.com/agenda/agenda/pull/464/commits/6ca2a3babe0de191437ff5055397d9c731e60254#diff-65704758a2c4406e3c56f4ca06c2518cR1186

nareshsr:
https://stackoverflow.com/questions/53460679/unhandledpromiserejectionwarning-when-using-agenda/63007932#63007932

PR #454, opened 2017-07-04
Update devDeps

OmgImAlexis:
After running `make test` I've noticed quite a few tests are taking around a second to process which often causes the tests to timeout. Maybe we could look at moving over to a test runner that runs all tests in parallel like [ava](https://github.com/avajs/ava) and/or increase the timeout?


simison:
LGTM

Seems to pass just fine with Travis and OK for adding yarn. @joeframbach is using Yarn with Agendash as well, right?

OmgImAlexis:
Yeah, [Looks like it](https://github.com/agenda/agendash/blob/master/yarn.lock).

joeframbach:
That is a very old yarn.lock file and I'm not tied to it.

PR #467, opened 2017-07-18
Add .editorconfig file, comment everything to help understand code

emhagman:
I added an `.edtiorconfig` file to standardize the formatting and also commented everything I possibly could to help people better understand the code base. I think this will greatly speed up development. Also, I changed some '==' to '===' where type coercion was not actually necessary.


lib/agenda.js
```diff
@@ -115,6 +116,14 @@ Agenda.prototype.db_init = function( collection, cb ){
   });
 };
 
+/**
+ * Internal method called if new indices have an error
+ * @param {Error} err error returned from index creation from before
+ * @param {*} result result passed in from earlier attempt of creating index
+ * @param {Agenda} self instance of Agenda
+ * @param {Function} cb called when indices fail or work
```
OmgImAlexis: Keep the same terminology, "fail or pass".

lib/agenda.js
```diff
@@ -136,51 +145,97 @@ function handleLegacyCreateIndex(err, result, self, cb){
   }
 }
 
+/**
+ * Set name of "queue" worker
```
OmgImAlexis: as "queue"

lib/agenda.js
```diff
@@ -415,108 +550,150 @@ Agenda.prototype.stop = function(cb) {
...
 Agenda.prototype._findAndLockNextJob = function(jobName, definition, cb) {
-  var self = this,
-      now = new Date(),
-      lockDeadline = new Date(Date.now().valueOf() - definition.lockLifetime);
 
-  // Don't try and access Mongo Db if we've lost connection to it. Also see clibu_automation.js db.on.close code. NF 29/04/2015
-  // Trying to resolve crash on Dev PC when it resumes from sleep.
+  var self = this, now = new Date(), lockDeadline = new Date(Date.now().valueOf() - definition.lockLifetime);
+
+  // Don't try and access Mongo Db if we've lost connection to it.
```
OmgImAlexis: MongoDB

lib/agenda.js
```diff
@@ -415,108 +550,150 @@ Agenda.prototype.stop = function(cb) {
...
+  // Trying to resolve crash on Dev PC when it resumes from sleep. NOTE: Does this still happen?
```
OmgImAlexis: We should confirm this is still an issue. Maybe open an issue about it and ref this commit+line.

lib/agenda.js
```diff
@@ -415,108 +550,150 @@ Agenda.prototype.stop = function(cb) {
...
+ * NOTE: May need to update what properties get set here, since job unlocking seems to fail
+ * @param {Function} done callback when jobs are unlocked
+ * @access private
+ * @returns {undefined}
+ */
 Agenda.prototype._unlockJobs = function(done) {
   var jobIds = this._lockedJobs.map(function(job) {
     return job.attrs._id;
   });
   this._collection.updateMany({_id: { $in: jobIds } }, { $set: { lockedAt: null } }, done);    // NF refactored .update() 22/04/2015
```
OmgImAlexis: Might as well remove the other "NF refactor" comments.

lib/agenda.js
```diff
@@ -549,23 +726,35 @@ function processJobs(extraJob) {
...
 
-      // insert the job to the queue at its prioritized position for processing
+      // Insert the job to the queue at its prioritized position for processing
       jobQueue.splice(jobIndex, 0, job);
+
     });
   }
 
+  /**
+   * Internal method that will lock a job and store it on Mongo
```
OmgImAlexis: MongoDB

lib/agenda.js
```diff
@@ -578,59 +767,90 @@ function processJobs(extraJob) {
...
 
+    // Query to run against collection to see if we need to lock it
     var criteria = {
       _id: job.attrs._id,
       lockedAt: null,
       nextRunAt: job.attrs.nextRunAt,
       disabled: { $ne: true }
     };
 
+    // Update / options for the Mongo query
```
OmgImAlexis: MongoDB

lib/agenda.js
```diff
@@ -578,59 +767,90 @@ function processJobs(extraJob) {
...
     var update = { $set: { lockedAt: now } };
     var options = { returnOriginal: false };
 
+    // Lock the job in Mongo!
```
OmgImAlexis: MongoDB

lib/job.js
```diff
@@ -36,7 +36,11 @@ var Job = module.exports = function Job(args) {
   this.attrs = attrs;
 };
 
-Job.prototype.toJSON = function() { // create a persistable Mongo object -RR
+/**
+ * Given a job, turn it into an object we can store in Mongo
```
OmgImAlexis: MongoDB

lib/job.js
```diff
@@ -231,30 +298,40 @@ Job.prototype.isRunning = function() {
...
+ * @param {Function} cb called after job is saved or errors
+ * @returns {exports} instance of job
+ */
 Job.prototype.save = function(cb) {
   this.agenda.saveJob(this, cb);
   return this;
 };
 
+/**
+ * Remove the job from Mongo
```
OmgImAlexis: MongoDB

lib/job.js
```diff
@@ -231,30 +298,40 @@ Job.prototype.isRunning = function() {
...
+ * @param {Function} cb called when job is removed from mongo
```
OmgImAlexis: MongoDB

emhagman:
@OmgImAlexis I think that should do it! Let me know if you need me to change anything else.

OmgImAlexis:
@emhagman I was going to mention the missed "MongoDB" but then noticed the sneaky commit at the end. 😛

simison:
Any reason why
```js
var Job = require('./job.js'), humanInterval = require('human-interval'), utils = require('util'),
  Emitter = require('events').EventEmitter;
```

instead of
```js
var Job = require('./job.js'), 
    humanInterval = require('human-interval'), 
    utils = require('util'),
    Emitter = require('events').EventEmitter;
```

? :-)

emhagman:
@simison Honestly, I wanted to completely remove comma separated declarations ;) But no, no reason, I probably should not have touched that line!

OmgImAlexis:
@simison I merged with that since all the others are now using the same style. I'd rather switch to xo if everyone's okay with it since it's way more strict.

simison:
>As a side note, I know the .eslintrc file was just added but that doesn't seem to handle cases like this.

Well it would if it would be configured to do so. ;-)

I never used `xo` but I'd be happy to see it in action here so I can learn about it! :+1:

I like this [from XO's readme](https://github.com/sindresorhus/xo):

>Never discuss code style on a pull request again!

:-D

lushc:
The comments starting with "NOTE:" and asking questions shouldn't have made it in IMO. My issue with comments that ask the reader is that they're already reading the source code for clarification, so if "we" for example don't know what the single job type is then how will they? 😃 

Also, a number of comments are (again, IMO) kinda redundant. For example, comparing two added in this PR:

```js
// If we have things to protect, set them in MongoDB using $setOnInsert
if (Object.keys(protect).length > 0) {
  update.$setOnInsert = protect;
}
```
```js
// We have a result from the above calls
// findAndModify() returns different results than insertOne() so check for that
var res = result.ops ? result.ops : result.value;
```

The latter gives useful context to something not immediately clear, while the former is just re-explaining code that's already quite self-documenting.

I hope I'm not coming across too negatively as a lot of these comments I would've liked to be there the first time I looked through the source code, but I think slimming down the number of comments by removing questions and redundant explanations would improve readability further 👍

emhagman:
@lushc Of course not. I would like to remove these but at the same time I think the repo is currently in a weird state where there is old code that I am not sure even some of the current contributors know where it came from?

In the future I can submit an issue asking a question such as "What is the 'single' job type?", would that be the better way to handle this?

OmgImAlexis:
@lushc as @emhagman stated the current code base is a mix of old and new code so I merged on the basis that between all the new devs we probably don't know all of the code and as we all get more use to it they can be removed. The comments also help with refactoring and linting since it shows from a user's point of view what does what without necessarily reading the full section of code.

I'm 100% for removing 20% - 30% of the comments that were added but not until we have this repo in a better state.

lushc:
@emhagman @OmgImAlexis thanks for clarifying, much appreciated! Regarding the questions, the easiest way would probably have been to list them in the PR initially and then @'d agenda/maintainers to get some answers. But of course feel free to open an issue as there may be gaps in the docs, especially when it comes to the internals.

Issue #358, opened 2016-08-30
Bug in calculating next month


mgregier:
My CRON expression looks like this: '0 12 4 8 *' ( 4th August at 12:00) but i can see that nextRunAt value in mongo is set to one month forward ( 4th September at 12:00)
I'm using job.computeNextRunAt(); function to calculate next run.
Internally job.js is using computeFromInterval funciton and is calling cronTime._getNextDateFrom(lastRun); to calculate next run. 
_getNextDateFrom is using date from moment.js which has 0-based months - January is 0 and in CRON expressions January is 1. That's why it is always calculating next run one month forward.

OmgImAlexis:
@mgregier Is this still an issue?

mgregier:
@OmgImAlexis I'm not sure because I'm not using Agenda anymore in my project. This was a bug that moment.js treated months as 0-based while in CRON expressions January has number 1 in months sequence.

ldsmike88:
Why is this still an issue?

leonardlin:
merged https://github.com/agenda/agenda/pull/1150

PR #475, opened 2017-07-19
Add test for bug #358, switch to cron-parser and remove cron

emhagman:
This closes bug #358 which is in fact, still a bug. In the docs for `node-cron`, it specifically says that the months are range `0-11` but the standard UNIX way to parse cron is `1-12`.

I have added a test that confirms this is still an issue and the test now passes.

Previously lib parsing: https://github.com/kelektiv/node-cron#cron-ranges
New lib parsing: https://github.com/harrisiirak/cron-parser#supported-format
Standard UNIX Cron: https://en.wikipedia.org/wiki/Cron

```
* * * * * *
| | | | | | 
| | | | | +-- Year              (range: 1900-3000)
| | | | +---- Day of the Week   (range: 1-7, 1 standing for Monday)
| | | +------ Month of the Year (range: 1-12)
| | +-------- Day of the Month  (range: 1-31)
| +---------- Hour              (range: 0-23)
+------------ Minute            (range: 0-59)
```

Hopefully not too many people are using `cron` to schedule jobs, as the month parsing is currently broken in the standards sense.

### DIFF
--- a/lib/job.js
+++ b/lib/job.js
@@ -3,7 +3,7 @@
  * None
  */
 
-var humanInterval = require('human-interval'), CronTime = require('cron').CronTime, date = require('date.js'),
+var humanInterval = require('human-interval'), CronTime = require('cron-parser').CronTime, date = require('date.js'),
     moment = require('moment-timezone'), debug = require('debug')('agenda:job');
 
 var Job = module.exports = function Job(args) {
--- a/lib/job.js
+++ b/lib/job.js
@@ -74,14 +74,6 @@ Job.prototype.computeNextRunAt = function() {
   }
   return this;
 
-  function dateForTimezone(d) {
-    d = moment(d);
-    if (timezone) {
-        d.tz(timezone);
-    }
-    return d;
-  }
-
   /**
    * Internal method that computes the interval
    * @returns {undefined}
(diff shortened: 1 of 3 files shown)


joeframbach:
Broken or not, it is a backwards-incompatible change and should be rolled out very carefully. It should be a major version bump with several deprecation warnings etc.

emhagman:
@joeframbach Completely agree. How do we currently go about scheduling versions?

simison:
>EDIT: For that matter, how should we be handling deprecation? I am not sure if this library has had to deal with that yet.

Good question. 

I added this to 1.0.0 milestone anyways.

Normally we could create a new method and keep+deprecate old `repeatEvery()` around for a while, but in this case I don't think that's smart.

I guess we just have to document well with changelog that this is a major change and add instructions how to update existing apps.

emhagman:
I think a changelog.md file with per version changes is a must, good idea. Thanks for adding the milestone!

_Sent from my Google Nexus 6P using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

joeframbach:
A separate docs cr would be nice.

emhagman:
@joeframbach I have added a PR at #481 to document the current difference in parsing, if you could take a loo that'd be great!

joeframbach:
Doc pr looks good. I merged it. Let's try to remember to merge this before 1.0.0

simison:
@emhagman This was failing due `cron` package being switched to `cron-parser` and thus `require('cron')` didn't work, trying to see if my small change will get tests pass.

Feel free to modify, I'd like to get this in soon together with other breaking changes.

simison:
@joeframbach @emhagman FYI I added docs change to [changelog](https://github.com/agenda/agenda/blob/master/History.md) as well

OmgImAlexis:
If we do switch to this lib we should keep an eye on https://github.com/harrisiirak/cron-parser/pull/109 since I don't really want to be using a cron lib that adds support for non-spec methods without a way to disable it.

harrisiirak:
Hello, I'm owner/maintainer of https://github.com/harrisiirak/cron-parser. @OmgImAlexis not sure where you take the idea that this PR will be merged. As you can see then the author of this PR is ignoring my request for adding some basic test cases and documentation. No test cases = no love :( Even if I'm start considering merging this branch, this is not going to be standard feature.

OmgImAlexis:
@harrisiirak that's why I said we should keep an eye on the PR to see what the outcome is. We already have a PR or two using your lib as a replacement for another one we were using. Before we merge them I'd like to check the PR I linked to see what happened.

cruisercohen:
I updated to agenda@1.0.3 and I'm still seeing 0 based months. I expected per the readme that after 1.0.0 the months would be 1 based?
As an example I'm doing 
`job.repeatEvery('35 18 9 1 *');`
`job.save();`
and my nextRunAt comes out to 2/9/2018, 6:35:00 PM

Sorry if I'm confused about something obvious.

simison:
Closing in favor of https://github.com/agenda/agenda/pull/1150

PR #479, opened 2017-07-21
Run Travis tests for multiple MongoDB versions

simison:
Repeating same environment variable causes TravisCI to run tests with both MongoDB versions on each NodeJS environment.

I'm not sure if this is too much testing but hey one can never be too careful, especially since MongoDB tends to move kinda fast and deprecate features often.

Docs: https://docs.travis-ci.com/user/environment-variables/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,16 +20,16 @@ env:
   global:
     # NodeJS v4+ requires gcc 4.8+
     - CXX=g++-4.9 CC=gcc-4.9
-script: make test-once
-env:
   matrix:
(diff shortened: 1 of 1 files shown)


simison:
@agenda/maintainers do you recon this would be good addition to Travis?

So it would test with both Mongo v2 and Mongo v3 for each Node version listed.
image

WDYT?

lushc:
Seems like an excellent idea to me 👍

niftylettuce:
Sounds great to me.

simison:
Cheers! :thumbsup: I'll merge.

PR #488, opened 2017-07-26
Clean/tests

OmgImAlexis:
I've changed over all vars to either let or const, there may still be a few vars but they're likely there because of some weird var hoisting rule since we have some functions that are called within themselves.

The `function()` calls were replaced with fat arrows.

The top level describe blocks were removed from the last test PR I merged so now each block is only called by it's filename.

### DIFF
--- a/test/fixtures/agenda-instance.js
+++ b/test/fixtures/agenda-instance.js
@@ -1,28 +1,30 @@
-var connStr = process.argv[2];
-var tests = process.argv.slice(3);
+const Agenda = require('../../index');
+const addTests = require('./addTests');
 
-var path = require('path'),
-    Agenda = require( path.join(__dirname, '..', '..', 'index.js' ) ),
-    addTests = require( path.join(__dirname, 'addTests.js') );
+const connStr = process.argv[2];
+const tests = process.argv.slice(3);
 
-var agenda = new Agenda({ db: { address: connStr } }, function(err, collection) {
-
-	tests.forEach(function(test) {
+const agenda = new Agenda({
+  db: {
+    address: connStr
+  }
+}, (err, collection) => {
+	tests.forEach(test => {
 	  addTests[test](agenda);
 	});
 
 	agenda.start();
 
 	// Ensure we can shut down the process from tests
-	process.on('message', function(msg) {
-	  if( msg == 'exit' ) process.exit(0);
+	process.on('message', msg => {
+	  if(msg === 'exit') {
+      process.exit(0);
+    }
 	});
 
 	// Send default message of "notRan" after 400ms
-	setTimeout(function() {
+	setTimeout(() => {
 	  process.send('notRan');
 	  process.exit(0);
 	}, 400);
-
 });
-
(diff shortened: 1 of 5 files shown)


OmgImAlexis:
@simison would you mind reviewing this as well please.

test/fixtures/agenda-instance.js
```diff
@@ -1,28 +1,30 @@
...
-var tests = process.argv.slice(3);
-
-var path = require('path'),
-    Agenda = require( path.join(__dirname, '..', '..', 'index.js' ) ),
-    addTests = require( path.join(__dirname, 'addTests.js') );
-
-var agenda = new Agenda({ db: { address: connStr } }, function(err, collection) {
-
-	tests.forEach(function(test) {
+const Agenda = require('../../index');
```
OmgImAlexis: The reason I switched to this is if we want to move to ES6 with babel in the future to allow for tree shaking we would likely end up using `import x from '../x';` and you can't use dynamic file paths with imports. Also it's for the most part considered a bad idea to use dynamic requires like that.

Maybe we should open an issue for this discussion so we can add an eslint plugin to force the chosen way?

simison: Aight, valid point! Seems like good to me like it is now as a transition towards XO — how was it, _Never discuss code style on a pull request again!_ :-D

simison:
I skimmed trough and looks good, but unfortunately I'm short on time today for more in depth look. I'll have more changes over the weekend!

test/retry.js
```diff
@@ -52,43 +43,41 @@ describe('agenda', function() {
...
 
-      jobs.now('a job');
+    jobs.now('a job');
 
-      jobs.start();
-    });
+    jobs.start();
   });
-
 });
```
simison: Doesn't XO add extra empty line here?

test/retry.js
```diff
@@ -52,43 +43,41 @@ describe('agenda', function() {
...
+    });
 
-      jobs.on('fail:a job', function (err, job) {
-        job
-          .schedule('now')
-          .save();
-      });
+    jobs.on('fail:a job', (err, job) => {
+      if (err) {
+        // Do nothing as this is expected to fail.
```
simison: Not in scope for this PR, but as a food for thought for future, we should test this kind of stuff with:
```js
expect(err).not.to.be(undefined);
```

(Tho Expect.js probably has some better ways to test error objects)

And that said, we should probably move away from Expect.js as it wasn't maintained in ages anymore.

OmgImAlexis: My thoughts exactly.

I'll switch from mocha + expect.js + blanket to [avajs](https://github.com/avajs/ava) which will get us using the latest testing and coverage setup. It has something exactly for this.

test/retry.js
OmgImAlexis: Yep but the current `.editorconfig` removes it. I'll add them back once I start on the large xo changes.

lushc:
LGTM, very much looking forward to those xo and ava changes!

Issue #455, opened 2017-07-04
Better code linting


OmgImAlexis:
I noticed that eslint isn't a dev dep and that eslint isn't actually setup to lint any files even though we have a `.eslintrc`. Personally I'd go for switching to [xo](https://github.com/sindresorhus/xo) which is built on top of eslint and works out of the box without needing any config.

Either way we should turn on linting for the travis build.

PING @agenda/maintainers

simison:
Ready for 1.0.0 with this?

OmgImAlexis:
Yep I'm good with releasing `1.0.0` now if you want?

[History](https://github.com/agenda/agenda/blob/master/History.md) looks good according to the [milestone](https://github.com/agenda/agenda/milestone/4).

simison:
Yep good for me!

After that Agendash might need some slight refactoring.

PR #492, opened 2017-07-29
Add xo

OmgImAlexis:
Since we're starting on 1.0.0 I thought I'd get this done before we merge anything else so we start with a fully linted repo.

Currently xo is setup to only lint the lib directory as I still need to work on the test section. I'd also like to address some of the issues I've currently disabled with `// eslint-disable-line`.

At some stage we should also look into moving the internal functions from the lib files to either their own file or up to the top of the files so they can be reused and tested by themselves.

Closes https://github.com/agenda/agenda/issues/455


OmgImAlexis:
I don't really use codeclimate but from what I can see the issues it's stating will be fixed later on as I want to make sure these changes don't break anything before I change more.

OmgImAlexis:
I'm not 100% sure what I changed that broke the tests I'm looking into it though.

I'm trying to find the issue.
https://gist.github.com/OmgImAlexis/fe91daafc071189227f6adae308721eb

```diff
diff --git a/test/agenda.js b/test/agenda.js
index 1bbd91a..23b2b6b 100644
--- a/test/agenda.js
+++ b/test/agenda.js
@@ -305,11 +305,15 @@ describe('Agenda', () => {
 
       it('runs the job immediately', done => {
         jobs.define('immediateJob', job => {
+          console.log('running.');
           expect(job.isRunning()).to.be(true);
           jobs.stop(done);
         });
+        console.log('setting to now.');
         jobs.now('immediateJob');
+        console.log('starting.');
         jobs.start();
+        console.log('started.')
       });
     });
 
@@ -460,4 +464,4 @@ describe('Agenda', () => {
       });
     });
   });
-});
\ No newline at end of file
+});
```

test/job.js
```diff
@@ -672,7 +672,7 @@ describe('Job', () => {
 
       jobs.define('lock job', {
         lockLifetime: 50
-      }, () => {
+      }, (job, cb) => { // eslint-disable-line no-unused-vars
```
OmgImAlexis: @simison I think I found the reason this is failing. Even if two params aren't used they need to be there for this test to pass otherwise it only gets called once.

OmgImAlexis:
Since this doesn't actually change anything apart from adding in stricter linting I'm going to merge without a review. If anyone has an issue with this please let me know and we can always revert the commit.

simison:
@OmgImAlexis yup! When merging without review most important thing is just to merge via PR, so that things get tested properly.

simison:
And LGTM :-) Thanks!

OmgImAlexis:
Any chance you could ping rschmukler again about removing codeclimate? We're going to end up with a lot of "failed" PRs until that's removed. 😞

PR #500, opened 2017-08-10
fix master: add missing debug dependency

geekflyer:
@OmgImAlexis The current master of agenda@latest on npm is completely failing because the debug dependency is missing in package.json even though it is used throughout the code.

Please merge this urgently and republish :)

```
Error: Cannot find module 'debug'
    at Function.Module._resolveFilename (module.js:325:15)
    at Function.Module._load (module.js:276:25)
    at Module.require (module.js:353:17)
...

```

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,6 +33,7 @@
   "dependencies": {
     "cron": "~1.1.0",
     "date.js": "~0.3.1",
+    "debug": "^3.0.0",
     "human-interval": "~0.1.3",
     "moment-timezone": "^0.5.0",
     "mongodb": "^2.2.10"


OmgImAlexis:
@geekflyer published as `0.10.2`.

geekflyer:
@OmgImAlexis thx for quick resolving :)

simison:
@OmgImAlexis odd that Travis didn't fail on that?

geekflyer:
@simison that may be because `debug` is probably a transitive dependency of one of the devDependencies of agenda and in travis' test run the devDependencies are also installed. I think there are some npm packages to avoid this kind of problem by scanning the code and searching for used `require(` calls in the source code, but not sure if it's worth the effort.

geekflyer:
@OmgImAlexis @simison What I'm rather now wondering is why all those commits are gone from master?? Did you rewrite the master history after his incident? On npm now `0.10.2` is published but the package.json in master contains `0.10.0`.

simison:
@geekflyer just noticed the same, @OmgImAlexis any idea?

OmgImAlexis:
@geekflyer yep I wasn't paying attention to the branch I was in and published master to npm as `0.10.1`. I unpublished that version and pushed this PR as `0.10.2`. This is why I was suggesting at the start to use master and dev branches so backporting issues like this is a lot easier.

simison:
Valid point @OmgImAlexis. :-) Perhaps a quick discussion at maintainers thread #441 ?

PR #503, opened 2017-08-12
split lib/*.js into single files

OmgImAlexis:
I know this is another massive change but this was inevitable. The main file was getting way too large which meant every single time we have more than one person working on a section there's conflicts over the whole file. This fixes that as well as allowing us to setup debug more cleanly by each file having it's own debug field instead of every single function inside of `agenda.js` using `agenda:worker`.

@simison shall I just merge once the PR goes green? I'm more than happy to rebase any/all PRs that I've broken.


simison:
@OmgImAlexis Let's get quick a review for this.

While I really like the sentiment, I'm not sure if every method should have it's own file. I don't really have anything else to suggest tho so I'm giving this :thumbsup:, but would love to hear quick opinion from others.

If nobody else replies within ~24h I think we can merge.

Love to see this kind of big refactoring done!

lib/agenda/index.js
```diff
@@ -8,13 +8,15 @@ const Emitter = require('events').EventEmitter;
...
 class Agenda extends Emitter {
-  constructor(config = {}, cb) {
+  constructor(config, cb) {
     super();
 
     if (!(this instanceof Agenda)) {
       return new Agenda(config);
     }
 
+    config = config ? config : {};
```
OmgImAlexis: Node v4 and v5 are a pain. [Only 6+ supports default params](http://node.green/#ES2015-syntax-default-function-parameters).

simison: `config = config || {}`

OmgImAlexis:
I'm with you on that, the reason I went with all methods is that we might as well and it'll be cleaner seeing unit tests separated for each method.

For example you'd have `agenda/now.js` with it's test living in `test/agenda.now.js` completely separate from all other tests.

simison:
>it'll be cleaner seeing unit tests separated for each method.

This was my thinking as well for +1.

OmgImAlexis:
Once we merge this I'm going to switch all tests to avajs and split them up. That should help us get to the bottom of these flaky tests. Off to bed(4:30am here).

OmgImAlexis:
@simison would you mind reviewing the last commit.

simison:
LGTM!

Now that I think of this again, having everything split into small files will help to write atomic code.

Will be exciting to see ava tests in action!

Ready for merge?

PR #418, opened 2017-04-02
README to mention creating index for fields used by .unique()

code2be:
Hi,

Based on this issue: https://github.com/rschmukler/agenda/issues/232
Just thought modifying the README file to mention the need of creating index for fields used by `.unique()` method would be useful for many people.

Thanks.

Issue #406, opened 2017-02-15
Integrate `semantic-release`


adrukh:
Seeing the tension between people wanting to see releases published to npmjs, synced with github releases, etc., I'd like to suggest using [semantic-release](https://www.npmjs.com/package/semantic-release)!

### How does this help:
Using `semantic-release` greatly simplifies release processes. Once used, every commit to the `master` branch will publish a new build on npmjs.com and a new release on github.com.

### What does this need:
- Configuration setup by running `npm install -g semantic-release-cli; semantic-release-cli setup` once on a dev env of a maintainer of this wonderful project. npm token and github token will be captured and used as env vars on travis
- Commit messages need to follow semantic release convention (start with `fix: ...` for fixes, `feat: ...` for features, etc.)
- Giving up on manually deciding on release version numbers.

### What this actually does:
- adds `NPM_TOKEN` and `GH_TOKEN` as travis env variables
- adds `semantic-release` as a target to `package.json`
- adds `npm run semantic-release` to `.travis.yml` under the `after_success` section

Hope the maintainers of this wonderful project find this helpful! :)

OmgImAlexis:
Closing this as we already have a way to handle versions and it doesn't look like the other devs or any other uses want this at the moment.

PR #505, opened 2017-08-14
Update dependencies

simison:
- Update `cron` and `debug` packages to latest
- Set semver version range to more strict `~version` ("Approximately equivalent to version") instead of `^version` ("Compatible with version")

Note that `debug` has 3 breaking changes: https://github.com/visionmedia/debug/blob/master/CHANGELOG.md#300--2017-08-08

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,12 +33,12 @@
     "url": "https://github.com/agenda/agenda/issues"
   },
   "dependencies": {
-    "cron": "~1.1.0",
+    "cron": "~1.2.1",
     "date.js": "~0.3.1",
-    "debug": "^2.6.8",
+    "debug": "~3.0.0",
     "human-interval": "~0.1.3",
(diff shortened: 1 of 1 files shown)


simison:
@emhagman you know about `debug` right? 

Am I right that none of these affect us:
- Breaking: Remove DEBUG_FD (#406)
- Breaking: Use Date#toISOString() instead to Date#toUTCString() when output is not a TTY (#418)
- Breaking: Make millisecond timer namespace specific and allow 'always enabled' output (#408)

([via debug changelog](https://github.com/visionmedia/debug/blob/master/CHANGELOG.md#300--2017-08-08))

simison:
@emhagman WDYT, should I add another round of tests to Travis with debug turned on?

OmgImAlexis:
@simison checking the PRs for those show we aren't effected by any of the breaking changes.

LGTM.

PR #509, opened 2017-08-16
dont try and unlock jobs when _lockedJobs is empty

OmgImAlexis:
This removes a useless database call since the find statement will return nothing if an empty array is passed.

### DIFF
--- a/lib/agenda/stop.js
+++ b/lib/agenda/stop.js
@@ -18,8 +18,21 @@ module.exports = function(cb) {
   const _unlockJobs = function(done) {
     debug('Agenda._unlockJobs()');
     const jobIds = self._lockedJobs.map(job => job.attrs._id);
-    debug('about to unlock jobs with ids: %O', jobIds);
-    self._collection.updateMany({_id: {$in: jobIds}}, {$set: {lockedAt: null}}, done);
+
+    if (jobIds.length >= 1) {
+      debug('about to unlock jobs with ids: %O', jobIds);
+      self._collection.updateMany({_id: {$in: jobIds}}, {$set: {lockedAt: null}}, err => {
+        if (err) {
+          return done(err);
+        }
+
+        self._lockedJobs = [];
+        return done();
+      });
+    } else {
+      debug('no jobs to unlock');
+      return done();
+    }
   };
 
   debug('Agenda.stop called, clearing interval for processJobs()');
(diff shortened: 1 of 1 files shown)


lib/agenda/stop.js
```diff
@@ -18,8 +18,21 @@ module.exports = function(cb) {
   const _unlockJobs = function(done) {
     debug('Agenda._unlockJobs()');
     const jobIds = self._lockedJobs.map(job => job.attrs._id);
-    debug('about to unlock jobs with ids: %O', jobIds);
-    self._collection.updateMany({_id: {$in: jobIds}}, {$set: {lockedAt: null}}, done);
+
+    if (jobIds.length >= 1) {
```
simison: BTW for readability might be worth considering exiting early if you don't mind multiple return points:
```js
if (!jobIds.length) {
  debug('no jobs to unlock');
  return done();
}

// ...
return done();
```

But I'm quite indifferent regarding these.

lib/agenda/stop.js
```diff
@@ -18,8 +18,21 @@ module.exports = function(cb) {
...
+      debug('about to unlock jobs with ids: %O', jobIds);
+      self._collection.updateMany({_id: {$in: jobIds}}, {$set: {lockedAt: null}}, err => {
+        if (err) {
+          return done(err);
+        }
+
+        self._lockedJobs = [];
```
simison: Is there a test case for this change? (If it even is a change, but looks like previously it would just leave this array behind).

lib/agenda/stop.js
OmgImAlexis: Length should always be compared. If length is missing it'll return undefined. I'll remove the unneeded return though, added this after being up 48+ hours. Oops.

```console
> !undefined
true
```

lib/agenda/stop.js
OmgImAlexis: Before it'd just leave it so I just set it back to how it was when we instantiate Agenda which is `[]`.

I'll add a test to the add-ava branch.

lib/agenda/stop.js
simison: I meant that instead of but I'm just bikeshedding really. :-)

```
if jobs
 // unlock jobs
else
  // no jobs
```

could be
```
if !jobs
 // return no jobs

// unlock jobs
```

OmgImAlexis: Ah. Should be checking for errors before running application code. Switching.

Issue #497, opened 2017-08-08
Dropping support for MongoDB 2.x.x


OmgImAlexis:
According to MongoDB the end of life for v2.6 is October 2016 which means really we should only be supporting 3.x.x officially.

This would affect https://github.com/agenda/agenda/issues/490

PING: @agenda/maintainers

simison:
Sure!

simison:
Do we have anything we'd want to publish in between or should we jump directly to 2.x?

OmgImAlexis:
I think it'd be worth it to get xo and ava support added in `v1.1.0` before jumping to `v2.0.0`.
Then we can remove MongoDB support and publish `v2.0.0`.

simison:
@shellscape we'll push this in 2.x — reviewing #616 will speed things up! ;-)

PR #518, opened 2017-08-29
Remove testing for Mongo 2.x in Travis CI

simison:
Related to #497 "Dropping support for MongoDB 2.x.x"

>According to MongoDB the end of life for v2.6 is October 2016 which means really we should only be supporting 3.x.x officially.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,24 +12,20 @@ addons:
     sources:
       # gcc 4.8+ requires ubuntu-toolchain-r-test
       - ubuntu-toolchain-r-test
+      - mongodb-upstart
+      - mongodb-3.4-precise
     packages:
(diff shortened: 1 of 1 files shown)


simison:
Dropping 2.x support might still need changes to docs and probably some code cleanup, but I recon that could be in separate PR?

OmgImAlexis:
@simison let's get this merged as it's currently causing all builds to fail since travis is now using `3.4` by default and we try to test `3.2`.

If no one objects we'll merge this in 24 hours from this comment's date.

PING: @niftylettuce @simison @joeframbach @agenda/maintainers

OmgImAlexis:
@simison would you mind adding the docs for this when you get a chance.

simison:
Docs: https://github.com/agenda/agenda/pull/541

PR #532, opened 2017-09-30
Updated Readme.md with an example of promise

subodhpareek18:
Hi, thank you for this really useful library! I'm sure many of us would agree that Agenda should support promises out of the box, but before we reach such a stage, I just wanted to add a useful example that users could implement themselves. 

It's something I immediately struggled with as I am trying to set it up for the first time, with having little experience with callbacks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -669,6 +669,27 @@ job.save(function(err) {
 })
 ```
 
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
If this is to be accepted I'd suggest using promisify from util instead of wrapping it with a new promise.

subodhpareek18:
I did try that, but it didn't seem to work for me.

simison:
Agenda now supports promises natively :-)

PR #541, opened 2017-10-18
Update documentation to reflect Mongo v3

simison:
https://github.com/agenda/agenda/blob/master/lib/agenda/db-init.js#L40-L42

https://docs.mongodb.com/manual/reference/method/db.collection.createIndex/#db.collection.createIndex

I'm not sure if this should be done only on agenda ready event or directly at the custom collection before its passed to Agenda?

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Install via NPM
 
     npm install agenda
 
-You will also need a working [Mongo](https://www.mongodb.com/) database (2.6+) to point it to.
+You will also need a working [Mongo](https://www.mongodb.com/) database (v3) to point it to.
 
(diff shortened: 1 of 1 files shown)


OmgImAlexis:
I think we should be fine with on ready. If we get any users reporting issues we'll have a look more into this. Since this should trigger the background index building we should have no difference when we run it.

simison:
Aight, feel free to merge / approve-review!

simison:
BTW Creating index could be a public `agenda.createIndex()` method?

OmgImAlexis:
Sounds good to me.

PR #542, opened 2017-10-21
Helping switch to ava

C45tr0:
Hopefully this helps. 

Was going to look at the integration tests that are skipped currently also.

Also, noticed some occasional uncaught rejections. Seems to be popping up on the same two tests when it does. Was going to look at that some also.

### DIFF
--- a/test/helpers/index.js
+++ b/test/helpers/index.js
@@ -9,13 +9,22 @@ const debug = new Debug('agenda:test:helpers');
 const jobProcessor = () => {};
 
 const startAgenda = async agenda => {
-  return new Promise(resolve => {
-    agenda.on('ready', () => {
-      agenda.processEvery(500);
-      agenda.start();
-      debug('agenda started');
-      resolve();
-    });
+  const onReady = resolve => {
+    agenda.processEvery(500);
+    agenda.start();
+    debug('agenda started');
+    resolve();
+  };
+
+  await new Promise(resolve => {
+    if (agenda.isReady) {
+      onReady(resolve);
+    } else {
+      agenda.on('ready', () => {
+        agenda.isReady = true;
+        onReady(resolve);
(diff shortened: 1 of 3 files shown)


C45tr0:
I looked into the uncaught rejections. It was a false assumption that it was only happening on a couple of tests. After looking at it more, it could happen on any of the tests.

It looks to be caused by dropping the database and something trying to ensure an index is created after the database has already been dropped.

test/helpers/index.js
```diff
@@ -9,13 +9,22 @@ const debug = new Debug('agenda:test:helpers');
...
+    debug('agenda started');
+    resolve();
+  };
+
+  await new Promise(resolve => {
+    if (agenda.isReady) {
+      onReady(resolve);
+    } else {
+      agenda.on('ready', () => {
+        agenda.isReady = true;
```
OmgImAlexis: I'm not sure I like setting this on the agenda object since it should remain pure the whole time we're testing.

@simison thoughts?

test/job/start.js
```diff
@@ -32,59 +31,73 @@ test('starts/stops the job queue', t => {
...
-  await startAgenda(agenda);
+  await new Promise(resolve => {
+    job.save((err, job) => {
+      t.ifError(err);
+      resolve(job);
+    });
+  });
 
-  t.true(lastRunAt === undefined);
+  await new Promise(resolve => {
```
OmgImAlexis: I'm not sure I get the reason behind this since this more or less skips the actual test since it should be checking if `lastRunAt` was still undefined.

C45tr0: It looked like that check wasn't helpful. I can add it back in if that was incorrect.

My thoughts were that the job should just not run and the t.fail() that is inside defined job accomplishes that.

PR #543, opened 2017-10-26
Make create index configurable

### LINKED ISSUE #498: Make createIndex configurable
- Add the option `{ index: true }` on the Agenda initialisation function to run or skip creating index.
- Add instructions how to create your index for production (or even a script that calls agenda's method?)
- Make `{index: true}` as default and have `{index: false}` as an option. This allows backwards compatibility since all current installs create the indexes by default.

C45tr0:
Resolves #498.

I still need to add documentation, but wanted to get it up for feedback.

I added config.events object as well, otherwise you could not add a callback to onReady as it ran synchronously.

### DIFF
--- a/lib/agenda/create-index.js
+++ b/lib/agenda/create-index.js
@@ -1,0 +1,28 @@
+'use strict';
+const debug = require('debug')('agenda:createIndex');
+
+/**
+ * Setup and initialize the collection used to manage Jobs.
+ * @param {String} collection name or undefined for default 'agendaJobs'
+ * @param {Function} cb called when the db is initialized
+ * @returns {undefined}
+ */
+module.exports = function(cb) {
+  const self = this;
+  debug('attempting index creation');
+  this._collection.createIndex(this._indices, {
+    name: 'findAndLockNextJobIndex'
+  }, (err, result) => {
+    if (err) {
+      debug('index creation failed');
+      self.emit('error', err);
+    } else {
+      debug('index creation success');
+      self.emit('index:created', result);
+    }
+
+    if (cb) {
+      cb(err);
+    }
+  });
+};
--- a/lib/agenda/db-init.js
+++ b/lib/agenda/db-init.js
@@ -12,19 +12,22 @@ module.exports = function(collection, cb) {
   debug('init database collection using name [%s]', collection);
   this._collection = this._mdb.collection(collection || 'agendaJobs');
   debug('attempting index creation');
-  this._collection.createIndex(this._indices, {
-    name: 'findAndLockNextJobIndex'
-  }, (err, result) => {
-    if (err) {
-      debug('index creation failed');
-      self.emit('error', err);
-    } else {
-      debug('index creation success');
-      self.emit('ready');
-    }
+  if (this._createIndex) {
+    this.createIndex(err => {
+      if (!err) {
+        debug('index creation success');
+        self.emit('ready');
+      }
+
+      if (cb) {
+        cb(err, self._collection);
+      }
+    });
+  } else {
+    self.emit('ready');
 
     if (cb) {
-      cb(err, self._collection);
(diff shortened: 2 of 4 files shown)


cornacchia:
Any possibility this will be merged in a future release?

simison:
Needs a rebase after promises rewrite as well Mongo driver update: https://github.com/agenda/agenda/pull/616

simison:
Another solution for this in https://github.com/agenda/agenda/pull/575

lib/agenda/create-index.js
```diff
@@ -0,0 +1,28 @@
+'use strict';
+const debug = require('debug')('agenda:createIndex');
+
+/**
+ * Setup and initialize the collection used to manage Jobs.
+ * @param {String} collection name or undefined for default 'agendaJobs'
+ * @param {Function} cb called when the db is initialized
+ * @returns {undefined}
+ */
+module.exports = function(cb) {
```
wingsbob: We've got async-await now! 😁

lib/agenda/db-init.js
```diff
@@ -12,19 +12,22 @@ module.exports = function(collection, cb) {
   debug('init database collection using name [%s]', collection);
   this._collection = this._mdb.collection(collection || 'agendaJobs');
   debug('attempting index creation');
```
wingsbob: Slightly misleading 😂

lib/agenda/index.js
```diff
@@ -31,6 +32,14 @@ class Agenda extends Emitter {
     this._sort = config.sort || {nextRunAt: 1, priority: -1};
     this._indices = Object.assign({name: 1}, this._sort, {priority: -1, lockedAt: 1, nextRunAt: 1, disabled: 1});
 
+    if (config.events) {
```
wingsbob: This seems somewhat unrelated

C45tr0:
Does #575 Cover this? Or should I get this updated?

simison:
>Does #575 Cover this? Or should I get this updated?

I'd rather see this more focused PR to be rebased. :-)

I'm working on Mongo v3 update here: https://github.com/agenda/agenda/pull/616 (testing / reviews welcome)

lib/agenda/index.js
C45tr0: It's been awhile, and with new changes this may not be needed anymore. But with out this there was no way to guarantee that you could receive an on ready event callback when no index was created. As it was happening synchronously on agenda instance creation.

C45tr0: I will review and see if this is still needed or not.

lib/agenda/db-init.js
C45tr0: I'll update this to be more appropriate.

C45tr0:
Sounds good. I see if I can get this taken care of tonight.

test/agenda.js
```diff
@@ -168,6 +168,124 @@ describe('Agenda', () => {
...
+
+    it('can be called even if already called', done => {
+      const agenda = new Agenda({mongo});
+
+      let indexCreated = 0;
+
+      agenda.on('index:created', result => {
+        expect(result).to.not.be.empty();
+
+        if (++indexCreated > 1) {
```
OmgImAlexis: Don't use `++` here. All code within the if statement should be read-only, no writes please.

lib/agenda/index.js
OmgImAlexis: Even if this is needed it should be in another PR.

C45tr0: Created #626 for this.

simison:
This would need just a couple small changes to get it over the finish line. :-)

SensitiveMix:
Please merge this PR ! we need it.

cornacchia:
Greetings. Any chance this PR will be merged?
We really need this feature.

Issue #457, opened 2017-07-04
repeatEvery job using cron time will run job immediately


xurt75:
I have tried a cron type job using the job.repeatEvery(). Whatever I set the cron-time, the job will start immediately once, and then nextRunAt the correct cron schedule time. It is confuse to me. 

I have added line 128 this.computNextRunAt() in repeatEvery() function in lib/job.js as follows. Then it works ok! I am not sure about this. Could you give some comments on this issue?
```
124 Job.prototype.repeatEvery = function(interval, options) {
125   options = options || {};
126   this.attrs.repeatInterval = interval;
127   this.attrs.repeatTimezone = options.timezone ? options.timezone : null;
128   this.computeNextRunAt();
129   return this;
130 };
```

simison:
Similar / duplicate: #422 _Scheduling job with agenda.every method with human interval runs the job immediately_

nemozny:
Yes, this one is a real pain!

I found a workaround though (and sorry, I can't get new lines in code working):
TESTjob.repeatEvery("1 2 * * *", {timezone: 'America/Chicago'});   
TESTjob.fail("Workaround to have a timestamp");
TESTjob.save();

**and then**

agenda.define('TESTjob', function(job, done) {
   if (new Date() - job.attrs.failedAt < 60000) {
   console.log("Only less than 1 minute (" +Math.floor((new Date() - job.attrs.failedAt) / 1000)+ "s) since the first run, so exit now");
   return done();
}
}

Please indent at your convenience.

**This way if Agenda tries to execute the job immediately, you calculate the duration since the job.fail timestamp (job.attrs.failedAt(date) - in miliseconds, so 1 minute is 60 seconds * 1000 ms) and break the execution with done().
In case you have doubts about this breaking the real job.fail, then you can check against job.attrs.failReason(string) too.**

simison:
@nemozny would this work:
```js
const interval = '30 minutes';
const job = agenda
  .create('myJob')
  .schedule(`in ${interval}`)
  .repeatEvery(interval);
```
via https://github.com/Codevate/agenda/commit/7f0667d2a2120c8a4566e2d086d325e446e86ce8#commitcomment-22911598 by @lushc

nemozny:
@simison No, this gets also executed immediately. Although I used this code, but I see no difference here, except maybe "in interval", which I cannot use:

const interval = '1 2 * * *';  
const job = agenda  
  .create('myJob')  
  .schedule(interval)  
[...]

PR #550, opened 2017-11-07
Refering to issue #457

arihantdaga:
Refering to issue #457

### DIFF
--- a/lib/job/repeat-every.js
+++ b/lib/job/repeat-every.js
@@ -10,5 +10,6 @@ module.exports = function(interval, options) {
   options = options || {};
   this.attrs.repeatInterval = interval;
   this.attrs.repeatTimezone = options.timezone ? options.timezone : null;
+  this.computeNextRunAt();
   return this;
 };


simison:
Hey, great to see this. Very elegant!

This will be an API change compared to previous behaviour but I think it's a good one. (FYI @maintainers)

@arihantdaga would you like to look into tests as well?

arihantdaga:
@simison Build is failing due to some another reason - It says something like this -  

`Error: Timeout of 8000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves` 
And failing for NodeJs 6. I do not know why it's happening but it's not related to this commit. 
And i think this commit wont be an API Change, because I think this is how repeatEvery is expected to run actually, or am i getting anything wrong ?

OmgImAlexis:
@arihantdaga the tests are flaky and that's a known issue. I've restarted the node v6 one and will keep doing it until it passes. We're currently moving to ava to replace the mocha tests so that'll be fixed soon.

arihantdaga:
@OmgImAlexis .Cool, Thanks.. 
@simison Can we merge this now ?

OmgImAlexis:
@arihantdaga as @simison said we need tests before we add this as it'll effect more than just yourself.

PR #555, opened 2017-12-01
Calculate the next run date on your own

marcuzy:
Agenda is an awesome scheduling lib and we've decided to migrate our project to it. We use complex schedule rules with Later.js. Since agenda uses its own parsers to interprets intervals we cannot just migrate to it (it's impossible). I think it would be awesome to let users use their own parsers.

Feel free to use your own parser like later.js 
```js
var agenda = new Agenda({
  getNextRunAt(job) {
    // calc and return the next run date
  }
});

agenda.every(
  'Every 2 weeks on Mon, Fri at 07:07 pm', 
  'my task', 
  { ... }, 
  { timezone: 'America/Los_Angeles' }
);
```
Also it will work for `agenda.schedule`, `job.repeatEvery`, `job.repeatAt` and `job.schedule`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,8 @@
 
 - Minimal overhead. Agenda aims to keep its code base small.
 - Mongo backed persistence layer.
-- Scheduling with configurable priority, concurrency, and repeating
-- Scheduling via cron or human readable syntax.
+- Scheduling with configurable priority, concurrency, and repeating.
+- Scheduling via cron, human readable syntax or your own parser (ex. using later.js).
 - Event backed job queue that you can hook into.
 - [Agendash](https://github.com/agenda/agendash): optional standalone web-interface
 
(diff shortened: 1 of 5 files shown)


marcuzy:
Whether are three +1 enough to merge? :) If somebody has an idea how to improve the feature, it would be awesome.

OmgImAlexis:
This PR will be merged eventually we just have others that need to be pulled in before this.
Ref: https://github.com/agenda/agenda/pull/557

jimmywarting:
This would be a grate way for me to implement a own parser eg like: [rrule](https://github.com/jakubroztocil/rrule)

It supports recurrence, interval, until, start & end.

But i don't like the `get` word in the beginning... feels like you are retrieving something...
the `computeNextRunAt` feels more appropriate i think.

PR #534, opened 2017-10-04
Drop support for legacy `ensureIndex()`

paramaggarwal:
Correctly calls `createIndex` but then even though there is no error, it proceeds to call `ensureIndex` as well, which is a deprecated method.

Fixes #490.

PR #557, opened 2017-12-06
Promise rewrite

OmgImAlexis:
This will be the PR used to track the Promise rewrite of Agenda.

100% of this will be breaking so don't hesitate to make big changes if it'll allow Agenda to work better when we release. We're going for 95% code coverage at launch and 100% by 2 weeks after.

@niftylettuce @simison @vkarpov15 @agenda/maintainers

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,8 @@ language: node_js
 services: mongodb
 sudo: required
 node_js:
-  - "4"
-  - "5"
-  - "6"
-  - "7"
   - "8"
+  - "10"
 addons:
   apt:
     sources:
--- a/README.md
+++ b/README.md
@@ -534,10 +534,9 @@ job queues can grab them / they are unlocked should the job queue start again. H
 shutdown.
 
 ```js
-function graceful() {
-  agenda.stop(function() {
-    process.exit(0);
-  });
+async function graceful() {
+  await agenda.stop();
+  process.exit(0);
 }
 
 process.on('SIGTERM', graceful);
--- a/README.md
+++ b/README.md
@@ -55,13 +55,13 @@ agenda.define('delete old users', function(job, done) {
   User.remove({lastLogIn: { $lt: twoDaysAgo }}, done);
 });
 
-agenda.on('ready', function() {
+agenda.on('ready', async function() {
   agenda.every('3 minutes', 'delete old users');
 
   // Alternatively, you could also do:
   agenda.every('*/3 * * * *', 'delete old users');
 
-  agenda.start();
+  await agenda.start();
 });
 
 ```
@@ -77,17 +77,17 @@ agenda.define('send email report', {priority: 'high', concurrency: 10}, function
   }, done);
 });
 
-agenda.on('ready', function() {
+agenda.on('ready', async function() {
   agenda.schedule('in 20 minutes', 'send email report', {to: 'admin@example.com'});
-  agenda.start();
+  await agenda.start();
 });
 ```
 
 ```js
-agenda.on('ready', function() {
+agenda.on('ready', async function() {
   var weeklyReport = agenda.create('send email report', {to: 'another-guy@example.com'})
   weeklyReport.repeatEvery('1 week').save();
-  agenda.start();
+  await agenda.start();
 });
 ```
 
@@ -323,8 +323,8 @@ An instance of an agenda will emit the following events:
 - `error` - called when Agenda mongo connection process has thrown an error
 
 ```js
-agenda.on('ready', function() {
-  agenda.start();
+agenda.on('ready', async function() {
+  await agenda.start();
 });
 ```
 
--- a/README.md
+++ b/README.md
@@ -874,7 +874,7 @@ when no connection is available on each [process tick](#processeveryinterval), a
 instance without having to restart the application.
 
 However, if you are using an [existing Mongo client](#mongomongoclientinstance)
-you'll need to configure the `reconnectTries` and `reconnectInterval` [connection settings](http://mongodb.github.io/node-mongodb-native/2.2/reference/connecting/connection-settings/)
+you'll need to configure the `reconnectTries` and `reconnectInterval` [connection settings](http://mongodb.github.io/node-mongodb-native/3.0/reference/connecting/connection-settings/)
 manually, otherwise you'll find that Agenda will throw an error with the message "MongoDB connection is not recoverable,
 application restart required" if the connection cannot be recovered within 30 seconds.
 
--- a/README.md
+++ b/README.md
@@ -56,14 +56,14 @@ agenda.define('delete old users', function(job, done) {
   User.remove({lastLogIn: { $lt: twoDaysAgo }}, done);
 });
 
-agenda.on('ready', async function() {
+(async function() { // IIFE to give access to async/await
   agenda.every('3 minutes', 'delete old users');
 
   // Alternatively, you could also do:
   agenda.every('*/3 * * * *', 'delete old users');
 
   await agenda.start();
-});
+})();
 
 ```
 
@@ -78,18 +78,18 @@ agenda.define('send email report', {priority: 'high', concurrency: 10}, function
   }, done);
 });
 
-agenda.on('ready', async function() {
+(async function() {
   agenda.schedule('in 20 minutes', 'send email report', {to: 'admin@example.com'});
   await agenda.start();
-});
+})();
 ```
 
 ```js
-agenda.on('ready', async function() {
+(async function() {
   var weeklyReport = agenda.create('send email report', {to: 'another-guy@example.com'})
   weeklyReport.repeatEvery('1 week').save();
   await agenda.start();
-});
+})();
 ```
 
 # Full documentation
@@ -153,7 +153,9 @@ You can also specify it during instantiation.
 var agenda = new Agenda({db: { address: 'localhost:27017/agenda-test', collection: 'agendaJobs' }});
 ```
 
-Agenda will emit a `ready` event (see [Agenda Events](#agenda-events)) when properly connected to the database and it is safe to start using Agenda.
+Agenda will emit a `ready` event (see [Agenda Events](#agenda-events)) when properly connected to the database.
+It is safe to call `agenda.start()` without waiting for this event, as this is handled internally.
+If you're using the `db` options, or call `database`, then you may still need to listen for `ready` before saving jobs.
 
 ### mongo(mongoClientInstance)
 
@@ -165,24 +167,29 @@ Please note that this must be a *collection*. Also, you will want to run the fol
 afterwards to ensure the database has the proper indexes:
 
 ```js
-agenda.on('ready', () => {
-  agenda._collection.createIndex({
-    disabled: 1,
-    lockedAt: 1,
-    name: 1,
-    nextRunAt: 1,
-    priority: -1
-  }, {
-    name: 'findAndLockNextJobIndex'
-  }, (err) => {
-    if (err) {
-      console.log('Failed to create Agenda index!');
-      console.error(err);
-    } else {
-      console.log('Agenda index created.');
-    }
-  });
-});
+(async () => {
+  await agenda._ready;
+
+  try {
+    agenda._collection.createIndex({
+      disabled: 1,
+      lockedAt: 1,
+      name: 1,
+      nextRunAt: 1,
+      priority: -1
+    }, {
+      name: 'findAndLockNextJobIndex'
+    });
+  } catch (err) {
+    console.log('Failed to create Agenda index!');
+    console.error(err);
+
+    throw err;
+  }
+
+  console.log('Agenda index created.');
+
+})();
 ```
 
 You can also specify it during instantiation.
@@ -320,13 +327,13 @@ By default it is `{ nextRunAt: 1, priority: -1 }`, which obeys a first in first 
 
 An instance of an agenda will emit the following events:
 
-- `ready` - called when Agenda mongo connection is successfully opened
+- `ready` - called when Agenda mongo connection is successfully opened and indeces created.
+        If you're passing agenda an existing connection, ou shouldn't need to listen for this, as `agenda.start()` will not resolve until indeces have been created.
+        If you're using the `db` options, or call `database`, then you may still need to listen for `ready` before saving jobs. `agenda.start()` will still wait for the connection to be opened.
 - `error` - called when Agenda mongo connection process has thrown an error
 
 ```js
-agenda.on('ready', async function() {
   await agenda.start();
-});
 ```
 
 ## Defining Job Processors
@@ -940,10 +947,8 @@ jobTypes.forEach(function(type) {
   require('./lib/jobs/' + type)(agenda);
 })
 
-if(jobTypes.length) {
-  agenda.on('ready', function() {
-    agenda.start();
-  });
+if (jobTypes.length) {
+  agenda.start(); // Returns a promise, which should be handled appropriately
 }
 
 module.exports = agenda;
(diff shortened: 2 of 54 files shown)


OmgImAlexis:
In regards to milestones I don't know what we're going todo with this since `2.0.0` has https://github.com/agenda/agenda/pull/534 added to it. Maybe we need to goto `3.0.0` or release a `1.5.0`?

simison:
In general it's better not to be afraid to bump [major version](https://semver.org/) when changes require it.

lib/agenda/database.js
```diff
@@ -1,42 +1,34 @@
...
+    if (!url.match(/^mongodb:\/\/.*/)) {
+      url = 'mongodb://' + url;
+    }
 
-  collection = collection || 'agendaJobs';
-  options = Object.assign({autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery}, options);
-  MongoClient.connect(url, options, (error, db) => {
-    if (error) {
+    // @TODO: This should be the mongodb instance
+    this.ctx.mongo = await promisify(MongoClient.connect)(url, options).catch(err => {
```
OmgImAlexis: Anyone know why this would be returning undefined instead of the mongodb object?

.travis.yml
```diff
@@ -2,8 +2,6 @@ language: node_js
 services: mongodb
 sudo: required
 node_js:
-  - "4"
-  - "5"
```
simison: Let's add this in a separate PR?

OmgImAlexis: This whole branch will be replacing the old one so there isn't much point in that.

OmgImAlexis:
Looks like we may need to drop node 6 if we want async without babel.

simison:
>Looks like we may need to drop node 6 if we want async without babel.

Node 6 is still very popular and LTS [ends only April 2019](https://github.com/nodejs/Release).

Would it be more work to add Babel or perhaps something like [caolan/async](https://github.com/caolan/async)?

OmgImAlexis:
Looks like a lot of places will be jumping straight to v8 since it was released at the same time as v6 and they need to migrate from v4 anyways.

Should we vote?

simison:
Healthy community should rarely need to vote. ;-) I feel we can just weigh cons and pros?

Bottom line is; backwards compatibility should be on the way of getting work merged. Adding Babel or a polyfill might slow us down too much.

That said, I'm sure with popularity of Agenda there are folks who need Node 6 support. They might be just okay using old version for now but we'd put them in difficult place if we next week find a security issue, although upgrading 6 → 8 was fairly painless. So it's just something to consider.

You have the steering wheel so I'll leave it up to you. :-) You could even publish this as a beta and see if anyone complains back. ;-)

OmgImAlexis:
If we do end up using babel I'd like to ship with the smallest amount of transforms possible.

simison:
>If we do end up using babel I'd like to ship with the smallest amount of transforms possible.

Agreed. Maybe good if you work on this PR now as if it's Node 8 and then look back and re-consider once everything is working. You could add Node 6 into "allowed to fail" matrix without removing it from engines list.

.travis.yml
```diff
@@ -5,6 +5,10 @@ node_js:
   - "6"
   - "7"
   - "8"
+matrix:
+  allow_failures:
+  - node_js: 6
+  - node_js: 7
```
OmgImAlexis: @simison any ideas here? I haven't used `allow_failures` before.

simison: Looks good otherwise but versions above are strings `"6"` and in matrix they're integers, might be it fails because of that.

See e.g. https://github.com/meanjs/mean/blob/482c38c287ff31688f0c9efa4ad4331009beeb40/.travis.yml#L4-L11

simison: `fast_finish: true` is handy btw: https://blog.travis-ci.com/2013-11-27-fast-finishing-builds/

OmgImAlexis:
We should add docs about `__pre` being the correct way to update the `this.config` and `this.ctx` on events.

simison:
Bah, added crucial word to this. Somehow dropped it before:
>backwards compatibility should **not** be on the way of getting work merged. Adding Babel or a polyfill might slow us down too much.

:-D

OmgImAlexis:
We need to discuss `ctx` vs `config`.

I want `.ctx` to be used as an internal store for each instance of the queue/job.

Where as `.config` should be for the whole queue or all of that job.

So if you run the job multiple times you can change the config on it and all update. If you change the `.ctx` only that job would change. This would also be the same for the agenda instance where each queue would have a `.ctx` and a `.config` where the config would update all queues but the `.ctx` would be for each individual queue.

lib/agenda/cancel.js
```diff
@@ -4,20 +4,20 @@ const debug = require('debug')('agenda:cancel');
...
  *  @caller client code, Agenda.purge(), Job.remove()
- *  @returns {undefined}
+ *  @returns {Promise} resolved (numRemoved) when cancellation fails or passes
  */
-module.exports = function(query, cb) {
-  debug('attempting to cancel all Agenda jobs', query);
-  this._collection.deleteMany(query, (error, result) => {
-    if (cb) {
+module.exports = function(query) {
+  return new Promise((resolve, reject) => {
```
vkarpov15: Why create a new promise when [`deleteMany()` already returns one](http://mongodb.github.io/node-mongodb-native/2.2/api/Collection.html#deleteMany)?

lib/agenda/schedule.js
```diff
@@ -9,66 +9,50 @@ const debug = require('debug')('agenda:schedule');
...
-   * @param {Function} cb called when job persistence in MongoDB fails or passes
    * @returns {module.Job} instance of new job
    */
-  const createJob = (when, name, data, cb) => {
-    const job = self.create(name, data);
-    job.schedule(when);
-    job.save(cb);
-    return job;
+  const createJob = (when, name, data) => {
+    return new Promise(async resolve => {
```
vkarpov15: This `async` is unnecessary

README.md
```diff
@@ -55,13 +55,13 @@ agenda.define('delete old users', function(job, done) {
   User.remove({lastLogIn: { $lt: twoDaysAgo }}, done);
 });
 
-agenda.on('ready', function() {
+agenda.on('ready', async function() {
```
vkarpov15: Putting this `async` here feels weird. Perhaps instead of relying on `agenda.on('ready')` we should just bake that into `agenda.start()` so clients would just do

```javascript
await agenda.start();

agenda.every('3 minutes', 'delete old users');
// ...
```

lib/agenda/cancel.js
OmgImAlexis: `Result.n` isn't a promise though and that's what we're resolving.

README.md
OmgImAlexis: Oh good point. We don't really even need ready now that we return a promise from `agenda.start()`, right?

lib/agenda/cancel.js
vkarpov15: so you can do `return this._collection.deleteMany(query).then(res => res.result.n)`, no?

OmgImAlexis: How's this?

vkarpov15: Much better, just one detail: you need to `throw err;` instead of `return err;` in the `catch()` handler. The `catch()` handler must either throw an error or return a promise that rejects, otherwise the resulting promise will be resolved. Example:

```javascript
const p = new Promise((resolve, reject) => reject(new Error('woops')));

// The below message will print even though `catch()` handler returned an error
p.
  catch(error => error).
  then(() => console.log('Not an error anymore!'));
```

OmgImAlexis:
I would highly appreciate a review or two on this.

PING @agenda/maintainers

lib/agenda/every.js
```diff
@@ -3,80 +3,61 @@ const debug = require('debug')('agenda:every');
...
  * @param {Number} interval run every X interval
  * @param {*} names String or strings of jobs to schedule
  * @param {Object} data data to run for job
  * @param {Object} options options to run job for
- * @param {Function} cb called when schedule fails or passes
- * @returns {*} Job or jobs created
+ * @returns {Promise} Job/s created. Resolves when schedule fails or passes
  */
-module.exports = function(interval, names, data, options, cb) {
+module.exports = async function(interval, names, data, options) {
```
vkarpov15: Worth noting that this means the new agenda will require node >= 7.6.0. I don't think that's the right choice since Node 6 [won't be EOL-ed until April 2019](https://github.com/nodejs/Release) so it's a little early to drop support for it.

OmgImAlexis: We could always add an async pollyfill such as https://github.com/touskar/async-await-es7?

OmgImAlexis: Also keep in mind this will be pushed to version `3.0.0` so people can still use version `2.0.0` if they don't want promise support.

vkarpov15: I'd suggest we just use [co](https://www.npmjs.com/package/co), works with node v4 with no transpilers and all you need to do is replace `async function() { /* ... */ }` with `co.wrap(function*() { /* ... */ });` and `await` with `yield`.

OmgImAlexis: Since it's using yield doesn't that mean instead of using native async we'd be using generators? I thought they were a lot slower?

OmgImAlexis: I'm personally against using a wrapper library when we could just drop support for older node versions. Anything below 7 or more specifically 8 can just use the v2 branch with callbacks, etc.

I'm not being on another project where we just sit back and don't progress because a few places that use this work for *free* can't/won't upgrade.

vkarpov15: I haven't seen anything about them being noticeably slower, been using both co and async/await in prod for years, including with agenda 0.9. If you're ok with me making the changes to switch from async/await to co internally, I'll do it. I'd be all for dropping support for node 4 and 6 if it added significant developer overhead, except for in this case the effort necessary to support node 4 and 6 is so marginal it seems silly to not support it.

OmgImAlexis: @agenda/maintainers thoughts?

OmgImAlexis: If you're willing to look after any async(co) wrapper code for the foreseeable future I'm okay with it.We'll need to make sure to update the jsdoc then. Any ideas on how co is represented in a jsdoc?

vkarpov15: `const fn = co.wrap(function*(){});` -> `fn()` returns a promise

OmgImAlexis: @vkarpov15 join us on Slack so we can discuss.

OmgImAlexis:
@bonesoul here's the update. @wingsbob has managed to fix most of the tests and we're now just working on the last few.

```
job concurrency
  - should not block a job for concurrency of another job
  - should run jobs as first in first out (FIFO)
  - should run jobs as first in first out (FIFO) with respect to priority
  - should run higher priority jobs first
  - should support custom sort option
```

lib/agenda/locklimit.js
```diff
@@ -3,11 +3,13 @@ const debug = require('debug')('agenda:locklimit');
...
 /**
  * Set the default amount jobs that are allowed to be locked at one time (GLOBAL)
- * NOTE: Is this different than max concurrency?
+ * @name Agenda#locklimit
+ * @function
  * @param {Number} num Lock limit
  * @returns {exports} agenda instance
  */
 module.exports = function(num) {
+   // @NOTE: Is this different than max concurrency?
```
wingsbob: Yup!

This seems to be the single most confusing aspect of agenda (looking through issues etc.)

Agenda will lock jobs 1 by one, which means it will set `lockedAt` in mongoDB, create an instance of the `Job` class and cache it into the `_lockedJobs` array. This defaults to having no limit, but can be managed using `lockLimit`

Agenda will also pull jobs from `_lockedJobs` and into `_runningJobs`. These jobs are actively being worked on by user code, and this is limited by `maxConcurrency`.

OmgImAlexis:
@agenda/maintainers what're we going todo in terms of async and node 4, 5 and 6?

lib/agenda/locklimit.js
simison: Definitely good material for docs. 👆  Wanna PR? :-)

karthikiyengar:
Thanks for this :-). I'm already using this branch and it seems to work fine. Anything I can do to speed this up?

OmgImAlexis:
@karthikiyengar I think we're just waiting on a full review and the tests to be finished off. I've had to take some time off of this to focus on other things but if you want to take a crack at it I'm fine helping you along.

bonesoul:
just dump 4,5,6 please.

wingsbob:
One thing which isn't added is support for promise-returning job definition functions

simison:
> Are we planning on adding a transpile step?

We've been talking dropping Node.js 6 support.

vkarpov15:
TBH I'd just merge this and move on. Adding back support for Node.js v6.x would take 20 minutes, replace async/await with co and be done with it. This PR is a small but meaningful step in the right direction for this lib.

karthikiyengar:
So gentlemen/women, what is the consensus on this one? 

If `co` is a priority, I'd be happy to add it - although I'm personally against introducing a library dependency made redundant by language features. 

I'd really appreciate if we can hear a clear voice from the library owners/maintainers on what specifically is required to get this merged.

@wingsbob - Can you please clarify? I'm not sure if `define` needs to be a promise here unless `define(...).then(() => agenda.now(..))` is a valid use case. If it is indeed going to be a promise, what should it resolve to?

wingsbob:
If the job processing function you pass to define returns a promise, is what I meant to say (if that is any clearer)

simison:
We'll just make small updates to the `README.md` and try to get this merged over the weekend.

vkarpov15:
image

karthikiyengar:
Thank you so much guys 😁

README.md
```diff
@@ -320,13 +327,13 @@ By default it is `{ nextRunAt: 1, priority: -1 }`, which obeys a first in first
 
 An instance of an agenda will emit the following events:
 
-- `ready` - called when Agenda mongo connection is successfully opened
+- `ready` - called when Agenda mongo connection is successfully opened and indeces created.
+        If you're passing agenda an existing connection, ou shouldn't need to listen for this, as `agenda.start()` will not resolve until indeces have been created.
```
OmgImAlexis: `ou` should be `you`

OmgImAlexis: `indeces` should be `indices`

README.md
```diff
@@ -320,13 +327,13 @@ By default it is `{ nextRunAt: 1, priority: -1 }`, which obeys a first in first
...
+        If you're using the `db` options, or call `database`, then you may still need to listen for `ready` before saving jobs. `agenda.start()` will still wait for the connection to be opened.
```
OmgImAlexis: replace "listen for `ready`" with "listen for the `ready` event"

lib/agenda/schedule.js
```diff
@@ -3,32 +3,28 @@ const debug = require('debug')('agenda:schedule');
...
  * Schedule a job or jobs at a specific time
+ * @name Agenda#schedule
+ * @function
  * @param {String} when when the job gets run
- * @param {*} names array of job names to run
+ * @param {Array<String>} names array of job names to run
  * @param {Object} data data to send to job
- * @param {Function} cb called when schedule fails or passes
- * @returns {*} job or jobs created
+ * @returns {Promise<Job>} job or jobs created
```
OmgImAlexis: Would this not be a "promise with a Job" or "an array of promises each resolving a job"? I think that's why this was using a `*`.

PR #561, opened 2017-12-17
Remove throw statements from process-jobs

andylhansen:
The file lib/utils/process-jobs.js throws errors that can't be caught, since process-jobs is initiated by setInterval. This branch replaces causes agenda to emit error events instead of throwing an exception.

### DIFF
--- a/lib/utils/process-jobs.js
+++ b/lib/utils/process-jobs.js
@@ -160,7 +160,9 @@ module.exports = function(extraJob) {
     // Lock the job in MongoDB!
     self._collection.findOneAndUpdate(criteria, update, options, (err, resp) => {
       if (err) {
-        throw err;
+        debug('job lock failed while locking on the fly');
+        self.emit('error', err);
+        return lockOnTheFly();
       }
       // Did the "job" get locked? Create a job object and run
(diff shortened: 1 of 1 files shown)


Getitdan:
Why can't it be merged yet? It's a severe issue.

lib/utils/process-jobs.js
```diff
@@ -160,7 +160,9 @@ module.exports = function(extraJob) {
     // Lock the job in MongoDB!
     self._collection.findOneAndUpdate(criteria, update, options, (err, resp) => {
       if (err) {
-        throw err;
+        debug('job lock failed while locking on the fly');
+        self.emit('error', err);
+        return lockOnTheFly();
```
wingsbob: Why the retry?

simison: I removed this retry to keep this PR more aligned with the original behaviour.

If it's needed, we can loop back in a separate PR.

simison:
Removed the retry (https://github.com/agenda/agenda/pull/561/#discussion_r194517380) and rebased.

@wingsbob wanna give this another review? Sadly I don't have time to write tests for this now. :-/

PR #584, opened 2018-03-30
Add optional callback to done() for job processing

princjef:
Under the current behavior, there is no reliable way of determining the result of calling `done()` when processing a job. There are 'success', 'fail' and 'complete' events, but these have to be correlated back to the specific job for which `done()` was called and they are unreliable for determining failures to save the job when calling `done(err)`.

This PR adds an optional callback to the `done()` function when processing a job. If provided, the callback is called with the result of saving the job's updated state to the database. This can be used to add instrumentation/logging or additional error handling to job processing without the extra correlation required when listening to events.


princjef:
There appears to be a flaky test that is messing up CI:

"Job" -> "job concurrency" -> "should run jobs as first in first out (FIFO) with respect to priority"

PR #587, opened 2018-04-03
Fix flaky test for FIFO ordering with priority

princjef:
Several PRs have failing builds due to intermittent errors with this specific test. The flakiness in the test is due to the fact that the two jobs that are supposed to run in priority order are scheduled using separately-constructed dates. If the two dates end up differing by even 1ms, the test will fail (since processing order is by time, then priority).

This change make the test use the same date for each of the relevant jobs so that it passes reliably. It also wraps the final expectations in a try/catch so that failures will show the expectation failure, rather than a timeout.

### DIFF
--- a/lib/agenda/db-init.js
+++ b/lib/agenda/db-init.js
@@ -14,7 +14,7 @@ module.exports = function(collection, cb) {
   debug('attempting index creation');
   this._collection.createIndex(this._indices, {
     name: 'findAndLockNextJobIndex'
-  }, (err, result) => {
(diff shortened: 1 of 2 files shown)


wingsbob:
Nice spot! Looks like this change has already been rolled into the promise rewrite https://github.com/agenda/agenda/pull/557/files#diff-c0b8dd124a1f328b82346ff619a637cbR854 however, so gonna close this.

If you feel something's been missed, feel free to say so and we can take another look 😄

PR #591, opened 2018-04-06
Release job from processing on lock expiration

### LINKED ISSUE #589: Jobs are not removed from the running/locked jobs when they expire
I am trying to reconcile locks in agenda with the way I typically see them work in other systems. Namely, if the lock for a job expires while a worker is processing it, the worker should no longer be allowed to complete/fail processing of the job (via `done()` and `done(err)`) because it no longer owns the resource. Similarly, once expired, the job should no longer count against the overall concurrency or lock limit, since the current worker doesn't really own the processing of that job any more.

However, the current behavior does neither of these things. From what I can tell, the validity of the lock is only checked immediately before the job is handed to the worker for processing, but not any time after that. Likewise, jobs whose locks have expired continue to count against the concurrency and lock limits even though the processing of them should no longer be valid. The only way to remove them is to call `done()` or `done(err)` to finish processing, which can be dangerous if there is another worker processing the job after the expiration of the lock.

This combination leads to situations like the one described in https://github.com/agenda/agenda/issues/389. An expired job is continually delivered upon expiration of the lock (which is reasonable), but the previous deliveries are never cleared out. This makes the same job count against you multiple times even though there is really only one valid delivery of that job.

Is this expected behavior? If so, would it be possible to also have the ability to do something more like what I described at the beginning?

princjef:
Fixes #589 by tracking the expiration time of the job's lock during processing and releasing the job from the list of running/locked jobs if the lock expires. This allows the job processor to continue to pick up new jobs for processing without expired jobs subtracting from the number of concurrent jobs that can be processed. It also adds `expire` and `expire:job name` events to agenda so that consumers can detect this situation.

This change would be considered breaking for consumers that expect to be able to call `done()` or `done(err)` after the job's lock has expired. I'm not sure whether that is considered an official/supported use case though. I couldn't think of a reasonable way to address the concurrency issue/memory leak around this without changing that behavior. I'd be happy to hear if people have other ideas for how to accomplish this.


Getitdan:
Can someone approve this?
The error: "Timeout of 8000ms exceeded" randomly occurs on different Node versions in:
"should run jobs as first in first out (FIFO) with respect to priority" test.

princjef:
Updated to align with the promise-based rewrite of agenda

lib/job/run.js
```diff
@@ -50,6 +54,66 @@ module.exports = function() {
       resolve(self);
     };
 
+    const expireCallback = function() {
```
wingsbob: Somewhat concerned about this approach.

Since the job may well have begun at this point, preventing it from being "done" could mean the job is simply run multiple times.

For a once-a-month type job, this could be really quite strange.

Why not check the lock is still "valid" when agenda is about to run the job, and extend it at that point?

princjef: Correct me if I'm misunderstanding, but I think we're talking about two different scenarios:

 1. Agenda locks and holds the job before giving it to the processing function. If the lock is already expired when it is about to be handed over, preemptively remove it. If it is not, potentially renew the lock with `touch()`
 2. Agenda locks the job and hands it over to the processing function with most/all of the lock duration still left. The processing function takes too long for the specified lock duration and/or there is some issue when trying to call `job.touch()` on the database, causing the lock to expire.

This PR is really focused on scenario 2, whereas I think what you're describing is scenario 1.

The problem I'm trying to solve with scenario 2 is that for various reasons lock renewal can fail and the lock may end up expiring. Let's say we have processor A and processor B. Processor A starts processing a job, but for one reason or another its lock expires. As soon as the lock has expired, processor B picks up the job, leading to two different processors trying to work on the job at the same time.

When that situation happens, the safest thing to do is for processor A (with its now-expired lock) to stop processing the job because it no longer controls it. However, there's no way for processor A to release the job without also completing/failing. If `done` is not called, the job is never removed from the running/locked jobs list and eventually processor A will no longer be able to process new jobs due to old expired jobs hanging around.

While this approach is more restrictive in not allowing you to complete the job if it is expired, it respects the locking semantics more closely in that it only allows you to process the job if you have an active lock on it.

wingsbob: I have a feeling I may have misunderstood the behaviour of `expireCallback`.

I expected it to prevent the `done` callback from being called, which, on re-reading, it doesn't seem to do.

It might be worth stressing the `expired` event(s) a bit more in the docs for people who might run into this.

princjef: Ah yes you're correct. Sorry I was remembering that I had actually blocked `done()` from being called once the job was expired, but all this really does at the moment is:

 * Provide a mechanism whereby you can be informed that a job has expired (via the `expire` and `expire:<job>` events)
 * Free up room for another job to be processed once the job has expired.

I can definitely change it to also block `done()` from being called if we want. It's a more aggressive change and is technically breaking relative to the current behavior. However, it's also safer because it helps prevent concurrent modification/processing. There's also the question of how to surface the error to the user if they try to call `done()` after the lock expires. Using the `error` event seems to be the most reasonable t me but I'm open to suggestions.

wingsbob: I think not blocking `done()` is OK. I'm kinda wondering whether this should be opt-in behaviour, as my understanding of `lockTime` is that it should be much longer than the job is expected to run (it certainly defaults to this, and I believe the documentation implies this too). 

We should definitely document that the `expired` event does not mean the job has not been run, and that it's intended to allow a job processor to abort.

I dunno, this feels like it's altering the behaviour of agenda to support a specific use-case in terms of job processors, and possibly also supporting "not calling `done()", which I feel isn't ideal.

princjef: I'd be fine with making this opt-in for people who want it and can make the additions to the documentation for the expire event. The main downside of putting it behind an option is that certain events would only be fired for certain configurations, which I don't think exists as of now.

As for whether it's special case, that really comes down to what locking means in agenda. The existence of the `touch()` method implies that jobs that run for longer than a single lock duration are expected to exist (even if they're rare) and in a typical time-based locking system you are blocked from modifying a resource if you don't own the lock for that resource. Since agenda (rather than the database) owns the conceptual lock concept, it makes sense for agenda to at least provide the ability for clients to detect and respond to the fact that they no longer have the lock on the job, even if it is not blocked outright. This helps prevent concurrent modifications from multiple job processors (though it isn't foolproof due to clock skew/race conditions).

wingsbob: Ok.

I think the best thing would be to try and flesh out the use-case for the `expire` events in the docs.

I believe you feel `touch`ing a job before it's handed to the processor should be a separate PR, right?

(Sorry this has dragged on 😞 )

wootwoot1234:
Any updates on this?  I would love to use it.

wootwoot1234:
@OmgImAlexis can you kick travis again?  Will that help get this merged?

lib/agenda/index.js
```diff
@@ -40,6 +40,7 @@ class Agenda extends EventEmitter {
     this._definitions = {};
     this._runningJobs = [];
     this._lockedJobs = [];
+    this._lockExpirationTimers = [];
```
koresar: I believe an ES6 `Set` would be better here. It would simplify some code (see below). Like
`this._lockExpirationTimers.has(...)` and `this.this._lockExpirationTimers.delete(...)`.

lib/agenda/stop.js
```diff
@@ -25,6 +25,12 @@ module.exports = function() {
         return resolve();
       }
 
+      debug('clearing out job lock expiration timers');
+      self._lockExpirationTimers.forEach(timer => {
```
koresar: The `for-of` loop is better in all circumstances IMO. More readable. Better CPU and RAM performance.
```js
for (const timer of self._lockExpirationTimers) clearTimeout(timer);
self._lockExpirationTimers.clear();
```

lib/job/run.js
```diff
@@ -20,6 +20,7 @@ module.exports = function() {
     self.computeNextRunAt();
     await self.save();
 
+    let lockTimeout = null;
```
koresar: I'd love to see a comment explaining why this variable exists, it's purpose. I can't quite understand from the code. Sorry.

PR #600, opened 2018-05-09
Remove support for deprecated versions of node

caiquecastro:
Node v4 and v5 does not receive updates now


simison:
Thanks! Made the same change elsewhere: https://github.com/agenda/agenda/pull/600

PR #605, opened 2018-05-22
Update README.md to have correct priority mapping

iblamefish:
The map has `normal: 0`, but documentation was `default: 0` - updated docs.


simison:
Thank you!

PR #575, opened 2018-03-01
mongodb upgrade to 3.x, noIndexCheck flag

simllll:
Closes https://github.com/agenda/agenda/issues/497

This PR upgrades the underlying mongodb driver to 3.0.3. Warning: you have to specify the database now by adding: 

```js
db: {
       database: 'agenda-test'
}
```
to agenda config, otherwise "agenda" is used as default database.

Furthermore you can add noIndexCheck: true (under db config) in production, so agenda does not perform an ensureIndex query.
```js
db: {
       noIndexCheck: true
}
```

Besides that I removed a code part that somehow stopped agenda from working from time to time (due to the reason it didn't reconnect correclty) and fixed the tests to run with the new driver. I also added the patch #561 by andylansen.

PR #616, opened 2018-06-10
Update native mongodb driver to v3 from v2

### LINKED ISSUE #497: Dropping support for MongoDB 2.x.x

simison:
See mongodb 3.x.x changes: 
https://github.com/mongodb/node-mongodb-native/blob/HEAD/CHANGES_3.0.0.md

Based on @simllll's work on https://github.com/agenda/agenda/pull/575

Resolves #497

### DIFF
--- a/lib/agenda/database.js
+++ b/lib/agenda/database.js
@@ -7,7 +7,7 @@ const debug = require('debug')('agenda:database');
  * @name Agenda#database
  * @function
  * @param {String} url MongoDB server URI
- * @param {String} collection name of collection to use
+ * @param {String} collection name of collection to use. Defaults to `agendaJobs`
  * @param {Object} options options for connecting
  * @param {Function} cb callback of MongoDB connection
  * @returns {exports}
@@ -26,7 +26,8 @@ module.exports = function(url, collection, options, cb) {
 
   collection = collection || 'agendaJobs';
   options = Object.assign({autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery}, options);
-  MongoClient.connect(url, options, (error, db) => {
+  MongoClient.connect(url, options, (error, client) => {
+    const database = client.s.options.dbName || 'agenda';
     if (error) {
       debug('error connecting to MongoDB using collection: [%s]', collection);
       if (cb) {
@@ -37,7 +38,8 @@ module.exports = function(url, collection, options, cb) {
       return;
     }
     debug('successful connection to MongoDB using collection: [%s]', collection);
-    self._mdb = db;
+    self._db = client;
+    self._mdb = client.db(database);
     self.db_init(collection, cb);
   });
   return this;
--- a/lib/agenda/database.js
+++ b/lib/agenda/database.js
@@ -27,7 +27,6 @@ module.exports = function(url, collection, options, cb) {
   collection = collection || 'agendaJobs';
   options = Object.assign({autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery}, options);
   MongoClient.connect(url, options, (error, client) => {
-    const database = client.s.options.dbName || 'agenda';
     if (error) {
       debug('error connecting to MongoDB using collection: [%s]', collection);
       if (cb) {
@@ -39,7 +38,7 @@ module.exports = function(url, collection, options, cb) {
     }
     debug('successful connection to MongoDB using collection: [%s]', collection);
     self._db = client;
-    self._mdb = client.db(database);
+    self._mdb = client.db();
     self.db_init(collection, cb);
   });
   return this;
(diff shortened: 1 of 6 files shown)


lib/agenda/database.js
```diff
@@ -26,7 +26,8 @@ module.exports = function(url, collection, options, cb) {
 
   collection = collection || 'agendaJobs';
   options = Object.assign({autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery}, options);
-  MongoClient.connect(url, options, (error, db) => {
+  MongoClient.connect(url, options, (error, client) => {
+    const database = client.s.options.dbName || 'agenda';
```
simison: I'm not sure how much we can trust this.

test/job.js
```diff
@@ -13,15 +13,17 @@ const Job = require('../lib/job');
...
 // Create agenda instances
 let agenda = null;
-let mongo = null;
+let mongoDb = null;
+let mongoClient = null;
 
 const clearJobs = () => {
   return new Promise(resolve => {
-    mongo.collection('agendaJobs').remove({}, resolve);
+    mongoDb.collection('agendaJobs').deleteMany({}, resolve);
```
simison: https://docs.mongodb.com/manual/reference/method/db.collection.deleteMany/

simison:
Noting that DB config and connection should be abstracted in tests, but I'll leave that out from this PR just for clarity.

lib/agenda/database.js
wingsbob: Could parse the connection string? https://github.com/mongolab/mongodb-uri-node

test/job.js
wingsbob: Pretty sure mongo client returns promises if you don't pass a callback, then you'd also get the rejection, rather than resolving with error|null

http://mongodb.github.io/node-mongodb-native/3.0/api/Collection.html#~deleteWriteOpCallback

lib/agenda/database.js
simison: I like the sentiment but the URL string has seen some changes (and might do so in the future, too). I'm not too happy to rely on an external package which hasn't been updated in years.

Tests should start failing if something changes with that object, so I'd be confident enough to keep it.

A viable alternative would be to require manually configuring it and now would be a great time to make that kind of breaking change.

test/job.js
simison: Hum, I'm getting unexpected test failure with this change:

image

lib/agenda/database.js
```diff
@@ -37,7 +38,8 @@ module.exports = function(url, collection, options, cb) {
       return;
     }
     debug('successful connection to MongoDB using collection: [%s]', collection);
-    self._mdb = db;
+    self._db = client;
+    self._mdb = client.db(database);
```
vkarpov15: I wouldn't reach in to `client.s.options.dbName`. If you do `client.db()` with no args, that gives you the database that is specified in the URI. [That's how mongoose does it](https://github.com/Automattic/mongoose/blob/fc76fe4f68ad013fcb7ca117e2539ee971a3679a/lib/connection.js#L493). Here are [the relevant docs](http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html#db)

package.json
```diff
@@ -42,7 +42,7 @@
     "debug": "~3.1.0",
     "human-interval": "~0.1.3",
     "moment-timezone": "~0.5.0",
-    "mongodb": "~2.2.33"
+    "mongodb": "~3.0.10"
```
vkarpov15: Let's bump to 3.1 since we need that for MongoDB 4 support

simison: No objections but I also think if it causes any extra work and slows down merging, it's smart to keep leave that for a separate PR.

lib/agenda/database.js
simison: >I wouldn't reach in to `client.s.options.dbName`. If you do `client.db()` ...

Excellent! I didn't like accessing it directly like that either.

simison:
> if I make the changes I suggested and then merge this PR?

Yes, please! 🙌

vkarpov15:
@niftylettuce @OmgImAlexis @simison I'm planning on merging this tomorrow afternoon unless anyone has any strong objections.

niftylettuce:
I'm open to anything with this project - recently switched to <https://github.com/bee-queue/bee-queue> until more progress is made to resolve core bugs and modernize this.  Wish I had more time to dedicate here, but love what you are doing. ❤️

simison:
Thank you @vkarpov15 !

fega:
this was already shipped to npm?

vkarpov15:
Nope I don't have publish access.

simison:
I've just published v2.0.0: https://github.com/agenda/agenda/blob/master/History.md#200--2018-07-19

niftylettuce:
@simison @OmgImAlexis you guys OK with giving @vkarpov15 NPM access?  He is a trusted member of the community and has supported Mongoose for years

simison:
@niftylettuce Absolutely! Thanks for asking. Consider done.

PR #626, opened 2018-06-13
Allow events to be setup inside config

C45tr0:
This allows users to set up all the event registration inside the initial agenda instance creation.


C45tr0:
Is this something that is still okay to add? Just wondering if we can get a yes or no on this soon.

PR #644, opened 2018-07-03
every: make the single type to be disabled

yorkie:
The every job is using the same document in the collection `jobs`, it might be configurable in user-land at some use cases.

Adding the `type` options for someone which wants every job to create a new instance(`job`) in every time :)

### DIFF
--- a/lib/agenda/every.js
+++ b/lib/agenda/every.js
@@ -23,7 +23,7 @@ module.exports = async function(interval, names, data, options) {
   const createJob = async (interval, name, data, options) => {
     const job = this.create(name, data);
 
-    job.attrs.type = 'single';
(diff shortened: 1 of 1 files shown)


yorkie:
The CI seems fail on the following unrelated problem:

```
$ sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install g++-4.9 gcc-4.9 mongodb-org-server mongodb-org-shell
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package g++-4.9
E: Couldn't find any package by glob 'g++-4.9'
E: Couldn't find any package by regex 'g++-4.9'
```

PR #650, opened 2018-07-13
Touch shouldn't accept a callback

pcorey:
Touch immediately calls `.save(cb)`, passing in the provided callback. `save` doesn't expect a callback, and throws a "no callback" error. `touch` should instead accept no callback and return a promise.


MichielDeMey:
Looks like this was fixed in https://github.com/agenda/agenda/pull/667. Thanks anyway! 👍

PR #652, opened 2018-07-19
Prevent unhandled promise rejections

princjef:
While working with the new promise code, I came across some code that results in unhandled promise rejections at the Node.js level for certain error cases. This PR addresses those and cleans up the promise patterns in a couple of other places to remove unnecessary calls to `new Promise()`.

In particular, the pattern below (`async` function in promise constructor) causes issues because the function provided to the promise constructor does not handle promises being returned (though it can handle synchronous exceptions being thrown)

```js
new Promise(async (resolve, reject) => {
    // Some code that may throw
});
```


simison:
Nice sleuthing!

> some code that results in unhandled promise rejections at the Node.js level for certain error cases

Do you recon any of those scenarios would need an additional test case? Though since failures might be cropping from MongoDB I could imagine adding tests not being very straightforward...

Code looks good. This will need a rebase before I'll give it a proper test spin.

princjef:
> Do you recon any of those scenarios would need an additional test case? Though since failures might be cropping from MongoDB I could imagine adding tests not being very straightforward...

Yeah that's exactly it. The primary source of asynchronous errors is the MongoDB calls themselves. For all of the non-test changes, the only way I can see to reproduce it would be to do some fault injection at the mongo level using spies and then trap `process.on('unhandledRejection')` to make sure it doesn't get triggered. I'd be willing to take a stab at it but I'm not sure it's worth the amount of test harness to set it up.

Code is rebased and should be ready for merging

PR #653, opened 2018-07-19
Support async functions in job processing

princjef:
Adds support for using `async` functions in asynchronous job processing in addition to the current `done` parameter that can be provided. The semantics are as follows:

 * When using a `done` param with no async function, the behavior is the same as before
 * When using an async function with no `done` parameter, the job is considered finished when the async function returns or the returned promise resolves (if using promises directly). If it throws or rejects the promise, the job is considered failed.
 * When both an async function and a `done` parameter are provided, either one can cause the job to fail (by calling `done(err)` or throwing from the async function), but only the `done()` function being called with no error can cause the job to succeed. This is mostly as a courtesy and avoids people having to wrap things in `new Promise()` just to prevent it from completing before they want it to.

cc: @OmgImAlexis

### DIFF
--- a/lib/job/run.js
+++ b/lib/job/run.js
@@ -18,7 +18,14 @@ module.exports = function() {
     self.computeNextRunAt();
     await self.save();
 
+    let finished = false;
     const jobCallback = async err => {
+      // We don't want to complete the job multiple times
+      if (finished) {
+        return;
+      }
+      finished = true;
+
       if (err) {
         self.fail(err);
       } else {
@@ -60,10 +67,10 @@ module.exports = function() {
       }
       if (definition.fn.length === 2) {
         debug('[%s:%s] process function being called', self.attrs.name, self.attrs._id);
-        definition.fn(self, jobCallback);
+        await definition.fn(self, jobCallback);
       } else {
         debug('[%s:%s] process function being called', self.attrs.name, self.attrs._id);
-        definition.fn(self);
+        await definition.fn(self);
         await jobCallback();
       }
     } catch (err) {
--- a/test/job.js
+++ b/test/job.js
@@ -328,6 +328,128 @@ describe('Job', () => {
       });
     });
 
+    it('allows async functions', async() => {
+      job.attrs.name = 'async';
+
+      const successSpy = sinon.stub();
+      let finished = false;
+
+      agenda.once('success:async', successSpy);
+
+      agenda.define('async', async() => {
+        await delay(5);
+        finished = true;
+      });
+
+      expect(finished).to.equal(false);
+      await job.run();
+      expect(successSpy.callCount).to.equal(1);
+      expect(finished).to.equal(true);
+    });
+
+    it('handles errors from async functions', async() => {
+      job.attrs.name = 'asyncFail';
+
+      const failSpy = sinon.stub();
+      const err = new Error('failure');
+
+      agenda.once('fail:asyncFail', failSpy);
+
+      agenda.define('asyncFail', async() => {
+        await delay(5);
+        throw err;
+      });
+
+      await job.run();
+      expect(failSpy.callCount).to.equal(1);
+      expect(failSpy.calledWith(err)).to.equal(true);
+    });
+
+    it('waits for the callback to be called even if the function is async', async() => {
+      job.attrs.name = 'asyncCb';
+
+      const successSpy = sinon.stub();
+      let finishedCb = false;
+
+      agenda.once('success:asyncCb', successSpy);
+
+      agenda.define('asyncCb', async(job, cb) => {
+        (async() => {
+          await delay(5);
+          finishedCb = true;
+          cb();
+        })();
+      });
+
+      await job.run();
+      expect(finishedCb).to.equal(true);
+      expect(successSpy.callCount).to.equal(1);
+    });
+
+    it('uses the callback error if the function is async and didn\'t reject', async() => {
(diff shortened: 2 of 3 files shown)


lib/job/run.js
```diff
@@ -60,11 +67,19 @@ module.exports = function() {
       }
       if (definition.fn.length === 2) {
         debug('[%s:%s] process function being called', self.attrs.name, self.attrs._id);
-        definition.fn(self, jobCallback);
+        try {
+          await definition.fn(self, jobCallback);
```
wingsbob: The arity check is to see whether the job processor function accepts a callback. I don't see any need to support accepting a callback *and* returning a promise

test/job.js
```diff
@@ -328,6 +328,128 @@ describe('Job', () => {
...
+          cb(err);
+        })();
+      });
+
+      await job.run();
+      expect(failSpy.callCount).to.equal(1);
+      expect(failSpy.calledWith(err)).to.equal(true);
+    });
+
+    it('favors the async function error over the callback error if it comes first', async () => {
```
wingsbob: This is the only test which fails without your changes, and as noted above I'm not sure about supporting use of both callbacks and promises in the same function

princjef: how were you running the tests? i get failures for the following tests when using the version of `lib/job/run.js` from master with the tests here:

 * allows async functions
 * handles errors from async functions
 * favors the async function error over the callback error if it comes first

in addition, this test technically passes, but throws an unhandled promise rejection on the process:

 * favors the callback error over the async function error if it comes first

princjef:
> I don't see any need to support accepting a callback and returning a promise

@wingsbob it's definitely a point up for debate and I ultimately think either allowing or explicitly forbidding it is a reasonable decision. Here are the two main reasons I opted for allowing both:

 * In places where people have the option of choosing one or the other, I occasionally see people use both at the same time. If we don't await it and their async function returns a rejected promise, they will get unhandled promise rejections and most likely will end up with hung jobs that never finish and gum up their system
 * In past experience with similar situations where each of these patterns are usually available on their own (like writing tests), I find that I sometimes have situations where I would like to be able to do something like the following and have errors handled for me automatically:

```js
it('does something', async done => {
  await agenda.start();
  await agenda.repeat('1 minute');
  agenda.define('some job', (job, cb) => {
    expect(job.attrs.name).to.equal('some job');
    done();
  });
});
```

If I can't use both patterns at the same time, the alternative choices are less elegant to write:

**Callbacks only:**

```js
it('does something', done => {
  (async () => {
    await agenda.start();
    await agenda.repeat('1 minute');
    agenda.define('some job', (job, cb) => {
      expect(job.attrs.name).to.equal('some job');
      done();
    });
  })().catch(done);
});
```

**Promises only:**

```js
it('does something', async () => {
  await agenda.start();
  await agenda.repeat('1 minute');
  await new Promise((resolve, reject) => {
    agenda.define('some job', (job, cb) => {
      expect(job.attrs.name).to.equal('some job');
      resolve();
    });
  });
});
```

For what it's worth, there are already places in the agenda tests that make an assumption that both patterns can be used at the same time. For the below example with the current code in master, if `agenda.stop()`, `clearJobs()` or the call to `agenda.define()` on line 456 throw, an unhandled promise rejection is created. In the case of `agenda.stop()` and `clearJobs()`, the throw will actually prevent `cb()` from being called, so the job will hang.

https://github.com/agenda/agenda/blob/24dbb4982a68f03485d266357aa1005d20c8a675/test/job.js#L452-L460

At the end of the day, I generally see two directions taken by libraries when a user may try to use mixed patterns like this. My preference is for the latter (I'm not particularly fond of mocha telling me how I can and can't write my tests), but both are definitely reasonable:

 * Throw an exception when both are detected (ideally on `agenda.define` in this case). you can check `fn.constructor.name === 'AsyncFunction'`, at declaration time, but a non-`async` function that returns a promise cannot be detected until it actually runs, so this is leaky.
 * Allow both and make sure that all forms of errors are handled gracefully. There is an implied semantic that when you have both, `done()` is how you complete successfully. That being said, this is typically the implicit expectation I see when people use it (as with the existing test shown above).

princjef:
Rebased with latest changes

lib/job/run.js
```diff
@@ -60,11 +67,19 @@ module.exports = function() {
...
+        } catch (err) {
+          await jobCallback(err);
+        }
       } else {
         debug('[%s:%s] process function being called', self.attrs.name, self.attrs._id);
-        definition.fn(self);
-        await jobCallback();
+        try {
```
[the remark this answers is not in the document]
princjef: good point. updated

simllll:
Oh, this isn't possible yet? Thought 2.0's goal was that everything can be used with promises now. Voting for this! Really would like to see this (especially the change is so small) in Agenda code base!

simllll:
Any news on this? After the new release I thought agenda will be alife again,...? :)

shc023:
Please land this 👍

simllll:
I'm still looking forward to this... is this still up?

wootwoot1234:
@wingsbob @princjef What's the status of this.  I would love to use it.

princjef:
Not sure. From what I can tell, this change should be compatible the changes made to the package since the PR was opened, so when it gets picked up by one of the maintainers it should be good to go as is.

derjp:
Please merge this, would love this feature. 👍

Issue #658, opened 2018-07-20
Fix linting warnings and errors


simison:
With package.json `engines` addition in https://github.com/agenda/agenda/pull/664, I'm getting some linting errors.
```
xo --verbose

  lib/agenda/save-job.js:8:4
  ⚠    8:4   Expected JSDoc for job but found err.  valid-jsdoc
  lib/agenda/start.js:5:1
  ⚠    5:1   Missing JSDoc return description.      valid-jsdoc
  lib/agenda/cancel.js:14:45
  ✖   14:45  Prefer await to then().                promise/prefer-await-to-then
  lib/utils/process-jobs.js:290:14
  ✖  290:14  Prefer await to then().                promise/prefer-await-to-then
  test/job.js:257:18
  ✖  257:18  Prefer await to then().                promise/prefer-await-to-then
  ✖  258:20  Prefer await to then().                promise/prefer-await-to-then
  ✖  322:22  Prefer await to then().                promise/prefer-await-to-then
  ✖  630:27  Prefer await to then().                promise/prefer-await-to-then
  ✖  640:21  Prefer await to then().                promise/prefer-await-to-then
  ✖  656:27  Prefer await to then().                promise/prefer-await-to-then
  ✖  802:64  Prefer await to then().                promise/prefer-await-to-then
  ✖  803:64  Prefer await to then().                promise/prefer-await-to-then
  ✖  804:68  Prefer await to then().                promise/prefer-await-to-then
  2 warnings
  11 errors
```

Warnings mentioned also at https://github.com/agenda/agenda/pull/656#issuecomment-406470642

PR #664, opened 2018-07-22
Add engines to package.json

simison:
Require Node.js 8 or newer. This is already a requirement by us so won't effectively change anything.

https://docs.npmjs.com/files/package.json#engines

https://jubianchi.github.io/semver-check/

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,9 @@
     "sinon": "6.1.4",
     "xo": "0.21.1"
   },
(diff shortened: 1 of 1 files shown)


simison:
Looks like this causes some linting errors (in addition to previous warnings):

https://github.com/agenda/agenda/issues/658

caiquecastro:
Duplicate of #689

PR #665, opened 2018-07-23
Allow users to sort and limit jobs when searching

### LINKED ISSUE #649: Suggestion: Add support for findOne/limit/sort rather than just find with the Agenda.Jobs() function
Currently I'm trying to find and modify the most recent version of a job _(all with the same name but different data)_ by looking for the most recent `lastRunAt` value. The `agenda.jobs()` function will return **all** jobs that match and doesn't allow sorting or limiting.

Could we add a function that lets you pass in a limit and sort query rather than just a find query please? It would be really helpful because at the moment I have to use `agenda._collection` to call `find()` myself which breaks the elegance of the library and the abstraction of the data layer. 

I am happy to take a stab this myself if you are open to random PRs. Thanks!

edwin-jones:
This commit resolves #649 

The code change is minimal but as this is my first contribution to your project I'd like to be sure I'm meeting your standards and the supplied test is in the correct location and format. Linter checks have passed as expected.

### DIFF
--- a/lib/agenda/jobs.js
+++ b/lib/agenda/jobs.js
@@ -6,10 +6,16 @@ const {createJob} = require('../utils');
  * @name Agenda#jobs
  * @function
  * @param {Object} query object for MongoDB
+ * @param {Object} sort object for MongoDB
+ * @param {Number} limit number of documents to return from MongoDB
  * @returns {Promise} resolves when fails or passes
  */
-module.exports = async function(query) {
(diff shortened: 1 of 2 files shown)


edwin-jones:
@simison this one has been pending for a long time...is it still wanted and or needed?

simison:
Would be good if you can follow up with documentation update after merging.

PR #667, opened 2018-07-25
Make job.touch() promise-based

princjef:
The [`job.touch()`](https://github.com/agenda/agenda/blob/master/lib/job/touch.js#L10-L13) method didn't get migrated over to promises along with everything else and still expects a callback. However, when it passes that callback down to `job.save()`, an error is thrown because a callback is not allowed.

This managed to slip through because the test for it changed at some point to mock out the implementation of `job.save()` with a callback version that didn't throw when it was called with a callback.

https://github.com/agenda/agenda/blob/bd8a8e003cd09d6e9826accbf6c30be75212a9a9/test/job.js#L352-L364

This PR makes the behavior of `touch()` align with the rest of the library and updates the associated test to remove the mock on `save()`.

Side note: this seems safe to treat as a bugfix rather than a breaking change because passing a callback to `touch()` doesn't work with the current code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -739,22 +739,21 @@ Disables the `job`. Upcoming runs won't execute.
 
 Enables the `job` if it got disabled before. Upcoming runs will execute.
 
-### touch(callback)
+### touch()
 
 Resets the lock on the job. Useful to indicate that the job hasn't timed out
-when you have very long running jobs.
+when you have very long running jobs. The call returns a promise that resolves
+when the job's lock has been renewed.
 
 ```js
 agenda.define('super long job', (job, done) => {
-  doSomeLongTask(() => {
-    job.touch(() => {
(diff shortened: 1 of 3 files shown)


simison:
Excellent, thanks for the fix! I [added this to changelog](https://github.com/agenda/agenda/commit/ff94c8a4c9bc564a0bed9eaa79de1c4fdbed0fde).

simison:
@princjef would you mind updating the docs before we merge?

https://github.com/agenda/agenda#touchcallback

princjef:
Docs updated

nanw1103:
Please publish an updated NPM with this fix. Otherwise the job.touch is totally broken. Thanks.

simison:
@nanw1103 Done with v2.0.1 :tada:

Issue #673, opened 2018-08-03
Can't connect Agenda with MongoDB Atlas


jayakumarp:
Seems like agenda appending "mongodb://". So, We could not connect agenda with MongoDB atlas. 

FYI:
lib -> agenda -> database.js

const self = this;
  if (!url.match(/^mongodb:\/\/.*/)) {
    url = 'mongodb://' + url;
  }

But the atlas connection string will be like
mongodb+srv://<username>:<password>@<mongo-cluster-url>/<database_name>

Please find the error below:

{ MongoError: Unescaped slash in userinfo section
    at parseConnectionString (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb-core/lib/uri_parser.js:235:21)
    at connect (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb/lib/operations/mongo_client_ops.js:179:3)
    at connectOp (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb/lib/operations/mongo_client_ops.js:283:3)
    at executeOperation (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb/lib/utils.js:420:24)
    at MongoClient.connect (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb/lib/mongo_client.js:168:10)
    at Function.MongoClient.connect (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/node_modules/mongodb/lib/mongo_client.js:371:22)
    at Agenda.module.exports [as database] (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/lib/agenda/database.js:31:15)
    at Agenda (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/node_modules/agenda/lib/agenda/index.js:56:12)
    at run (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/agenda.js:14:16)
    at Object.<anonymous> (/media/myname/DATA11/backup/Projects/myapp/code/apiserver/agenda.js:36:1)
  name: 'MongoParseError',
  message: 'Unescaped slash in userinfo section',
  [Symbol(mongoErrorContextSymbol)]: {} }

Tom5om:
[...]

PR #674, opened 2018-08-06
Changes to mongo atlas

Tom5om:
Update database.js to allow users to use a cluster with mongo atlas

### DIFF
--- a/lib/agenda/database.js
+++ b/lib/agenda/database.js
@@ -20,7 +20,7 @@ const debug = require('debug')('agenda:database');
  */
 module.exports = function(url, collection, options, cb) {
   const self = this;
-  if (!url.match(/^mongodb:\/\/.*/)) {
+  if (!url.match(/mongodb(?:\+srv)?:\/\/.*/)) {
     url = 'mongodb://' + url;
   }
 


simison:
Thank you! Could you please include a test or two to confirm the change in functionality? :-)

wingsbob:
Any chance someone could find a link to connection string spec/documentation? Would make it a lot easier to understand the change, especially as the regex includes a bonus `:` not mentioned in #673

simison:
https://docs.mongodb.com/manual/reference/connection-string/ <https://docs.mongodb.com/manual/reference/connection-string/>

shubhanshusingh:
please solve and merge this

Tom5om:
Hey @wingsbob and @simison I have updated the code, please review the PR

lib/agenda/check-url.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+/**
+ * Given a mongo connection url the function will check with a regex if it is valid
+ * @param url
+ * @returns {AggregationCursor|*|RegExpMatchArray|Promise<Response | undefined>}
```
wingsbob: I think this should return boolean?

Tom5om: This was automatically generated when I created the docblock, technically it doesnt return a boolean when it is true

wingsbob: Since this is now a separate function, I think it might be clearer to call it `needsMongoProtocol` and always return a boolean?

Tom5om: Aah I like that actually, good idea, as the function checks if it has the mongoProtocol, I have called it hasMongoProtocol

simison:
Shipped in Agenda v2.0.2

PR #680, opened 2018-08-16
agenda.jobs() usage update

verybluebot:
agenda.jobs() returns Promise instead of using a callback structure like in the docs.

changing the example to a Promise based one


simison:
Thanks!

Already updated in another commit: https://github.com/agenda/agenda/commit/a92d8360aec36a6671a2e632c7a867d9a6dcbb6c

PR #688, opened 2018-08-30
Adds more debugging logs to the job runtime

orta:
I'm struggling with `agenda.schedule`d being marked as ran inside agendash but not actually seeing the callback being called - so I thought I'd add a few more potential debugging hooks into agenda just in case.


simison:
Thanks!

Could you do a quick rebase with latest master, let's see them tests pass. :muscle:

orta:
I've rebased master, but the failures are still there 👍

orta:
I've submitted a PR that fixes the linter issues https://github.com/agenda/agenda/pull/701

PR #691, opened 2018-08-31
Cache yarn, test against latest node version

amilajack:


.travis.yml
```diff
@@ -3,7 +3,8 @@ services: mongodb
 sudo: required
 node_js:
   - "8"
-  - "10"
+  - node
+cache: yarn
```
OmgImAlexis: Please remove this, I've had nothing but issues with the yarn cache and it really doesn't speed anything up. We could switch to npm's ci mode if we really wanted faster installs.

.travis.yml
```diff
@@ -3,7 +3,8 @@ services: mongodb
 sudo: required
 node_js:
   - "8"
-  - "10"
+  - node
```
simison: If we're to add "latest", I'd rather have it in "allowed failures" not to have CI surprises slowing us down later on (kinda what we have now with MongoDB installs failing in CI).

OmgImAlexis: So maybe `LTS` and `latest`?

simison: I prefer clearly declared and stable CI but if you prefer to have `LTS` here, I'm not going to block it. :-)

If someday we're not maintaining this actively enough, the day when Agenda becomes incompatible with LTS, all CI build would start failing all of a sudden.

OmgImAlexis: My hope was once the ci build start failing that means we need to start migrating to the newer version, hopefully we're already there before it happens.

We could also do specific versions + latest instead?

simison: Yep, so it's mostly a taste-preference:
- don't declare version compatibility to "un-foreseen future": CI informs us when we're behind
- declare strict upper compatibility limit: danger to stay behind without CI informing us


Considering that our codebase is pretty cutting edge right now, I don't see huge risk doing either way. 👍  I'll leave it to you and others to consider. :-)

simison: >We could also do specific versions + latest instead?

I like this :thumbsup:

simison: Noting also that ~if~ when CI breaks, we can "fix" it by simply moving LTS to "allowed failures": we don't necessarily have to scramble to build support to get CI back green.

OmgImAlexis: You have me vote either way, so up to you.

MichielDeMey:
We’ve deprecated Node 8 support in `v3.0.0` and are now testing against the latest Node versions. 👍

PR #580, opened 2018-03-08
Added support for startDate and endDate in repeatEvery

designerkamal:
Based on #361 
```
job.repeatEvery('0 6 * * *', {
  startDate: new Date('2018-05-01T00:00:00.000Z'),
  endDate: new Date('2018-05-31T00:00:00.000Z')
});
// runs every day at 6pm from May 1 - May 15
job.save();
```
- Added tests
- Updated Readme

Also, added following rule to `xo`
```
"no-unused-vars": [
        "error",
        {
          "argsIgnorePattern": "^_"
        }
]
```
to get rid of that one linter error in `createIndex` callback. I am sure this will also be useful in the future in similar cases.

PR #720, opened 2018-10-15
Feature/start end dates

ganmor:
This is a rebase of #580 from @designerkamal

Since the pull request is getting old and we need this feature I thought I would help going forward

As a side node: tests do not pass since eslint do not pass
- Is it ok to disable the rule  promise/prefer-await-to-then

Also I noticed there is no eslint config: shall we create one ?


simison:
>Also I noticed there is no eslint config: shall we create one ?

Hmm, we use [`xo`](https://github.com/xojs/xo) — I wonder what's the preferred setup with xo and IDEs picking up eslint configs? Ping @OmgImAlexis 

https://github.com/agenda/agenda/blob/db30391881a887cc3da25b4828ab6ebe59d7fc63/package.json#L14

simison:
Noting that when merging this, would be nice to add Co-Authored-By tag so that @designerkamal gets kudos for his work. https://blog.github.com/2018-01-29-commit-together-with-co-authors/

leonardlin:
This has been implemented and is part of release 4.0.1

https://github.com/agenda/agenda/pull/361

PR #735, opened 2018-11-19
docs: fix wrong console log statement

stefanwalther:
- Console log should not contain single quotes but template literals


dandv:
Fixed by #771.

PR #752, opened 2019-01-02
docs(now): remove unsupported callback arg

vkarpov15:


simison:
Done elsewhere already, thanks!

PR #761, opened 2019-01-30
Fix JSDoc for .every: data & options are optional params

dandv:
Hope this trivial PR will be merged quickly.

### DIFF
--- a/lib/agenda/every.js
+++ b/lib/agenda/every.js
@@ -5,10 +5,10 @@ const debug = require('debug')('agenda:every');
  * Creates a scheduled job with given interval and name/names of the job to run
  * @name Agenda#every
  * @function
- * @param {Number} interval run every X interval
- * @param {*} names String or strings of jobs to schedule
- * @param {Object} data data to run for job
- * @param {Object} options options to run job for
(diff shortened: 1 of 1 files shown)


lib/agenda/every.js
```diff
@@ -5,10 +5,10 @@ const debug = require('debug')('agenda:every');
...
  * @name Agenda#every
  * @function
- * @param {Number} interval run every X interval
- * @param {*} names String or strings of jobs to schedule
- * @param {Object} data data to run for job
- * @param {Object} options options to run job for
+ * @param {String|Number} interval - run every X interval
+ * @param {*} names - String or strings of jobs to schedule
+ * @param {Object} [data] - data to run for job
+ * @param {Object} [options] - options to run job for
```
simison: What does `[]` signify here in this case? Shouldn't these be objects like

```js
{
  "key": "value",
  ...
}
```

dandv: The `[]` means that `data` and `options` are [optional parameters](http://usejsdoc.org/tags-param.html#optional-parameters-and-default-values) of type `object`.

PR #767, opened 2019-02-05
Fixed this={} in processDbResult

Solpatium:
`this` in `processDbResult` in `save-job.js` file is always `{}`, but it should be bound to agenda instance. Because of this a job is never run immediately, which is not a big issue, but it raises an error, for example, after Babel transpilation, because `this` is `undefined`. 

Fixing it breaks the tests, because 

```js
processJobs.call(this, job);
```

might actually run and it doesn't take `_jobsToLock` into consideration. This is why a check for this was added.


Beretta1979:
Could this be reviewed, need support for instant starting of jobs

Beretta1979:
Could this be reviewed , need this fix

mkaput:
rebased onto latest master

PR #793, opened 2019-03-28
Specify call order for start

dandv:
Calling `processEvery` after `start` means the interval will be ignored.

Calling `every` before `start` yields a Mongo Error:

> TypeError: Cannot read property 'findOneAndUpdate' of undefined
    at Agenda.module.exports [as saveJob]

### DIFF
--- a/README.md
+++ b/README.md
@@ -554,8 +554,8 @@ run them. You can also stop the queue.
 
 ### start
 
-Starts the job queue processing, checking `processEvery` time to see if there
-are new jobs.
+Starts the job queue processing, checking [`processEvery`](#processeveryinterval) time to see if there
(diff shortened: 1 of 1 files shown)


dandv:
@wingsbob See here for why `start` needs to be called before `every`.

wingsbob:
@dandv I believe the issue is specific to using `database` rather than `mongo`, as the former hands off creating the connection to agenda, where the latter requires the user to do that themselves.

Personally I feel that using `database` should be avoided, as there's no way to cleanly shut down. 😁 

The "more accurate" way to ensure that agenda is ready to schedule a job would be to either listen for the `ready` event, or await the `_ready` property.

Issue #781, opened 2019-03-12
run only one recurring job at a moment


mohitzim:
We can create different type of job using Agenda. I am using recurring job which is running at every 30 seconds. So every 30 seconds, a job is running. Sometimes a job is taking more than 30 seconds to run, but Agenda runs a scheduled job at 30 seconds (eventhough previous job is already running)

Following is the code :
```
const agenda = new Agenda({db: {address: db_address}, maxConcurrency: 1});

agenda.define('somejob', async (job, done) => {
  console.log('agenda job started..');
  somejob();
  done();
});
   
(async function() {
  await agenda.start();
  await agenda.every('30 seconds', 'somejob');
})();
```
Can we restrict agenda that if same type of job is running then dont run scheduled job ?

I tried with **maxConcurrency** and **defaultConcurrency** but didnt help. Can anyone please suggest solution ?

Thanks & Regards

dandv:
Actually Agenda does run only one recurring job at a moment! See the new [doc on `concurrency`](https://github.com/agenda/agenda/pull/795/files#diff-04c6e90faac2675aa89e2176d2eec7d8R286).

What I think happens in your case is that the job takes longer than the `lockLifetime` (default 10 minutes), so agenda assumes it might have crashed or timed out. See in the README,

> Typically you shouldn't have a job that runs for 10 minutes, so this is really insurance should the job queue crash before the job is unlocked.

To tell Agenda that your job is still running, call [touch()](https://github.com/agenda/agenda#touch) from the job processor.

If you still have problems, please try to reproduce based on the [example](https://github.com/agenda/agenda/pull/794) I've just added, and re-open the issue.

PR #794, opened 2019-03-29
Example illustrating concurrency and locking

dandv:
To help #781.

### DIFF
--- a/examples/concurrency.js
+++ b/examples/concurrency.js
@@ -1,0 +1,53 @@
+/**
+ * @file Illustrate concurrency and locking
+ */
+import Agenda from 'agenda';
+
+function time() {
+  return new Date().toTimeString().split(' ')[0];
+}
+
+function sleep(ms) {
+  return new Promise(resolve => setTimeout(resolve, ms));
+}
+
+const agenda = new Agenda({
+  db: {
+    address: 'mongodb://agendan:nuam0agenda@ds052408.mlab.com:52408/agenda',
+    options: { useNewUrlParser: true },
+    collection: `agendaJobs-${Math.random()}`,  // start fresh every time
+  },
+});
+
+let jobRuncount = 1;
+agenda.define('long-running job', {
+  lockLifetime: 5 * 1000,  // max amount of time the job should take
+  concurrency: 3,  // max number of job instances to run at the same time
+}, async (job, done) => {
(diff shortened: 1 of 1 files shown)


wingsbob:
Might be a bit clearer what the lock is about if you created more than 1 instance of `agenda`, and explained that only 1 instance would be able to work on the job?

Also, this doesn't look to have any example(s) for the concurrency setting(s)

dandv:
Happy to add those next, or in a separate example. This is the first example in the repo after all, and I'd like it to help folks like @mohitzim (see #781) understand how one instance Agenda works.

wingsbob:
Ok. I feel there's 2 different opinions on how locking should work.

The default behaviour of agenda is to lock jobs for much longer than it should take for them to process, and assume that if that expires, then agenda crashed while processing/with the job queued.

Your example kinda leans more towards having a very short lock time, and refreshing that lock in the job processor. This might cause some slightly odd behaviour, as agenda would still have the job in it's locked queue if the concurrency limit has been reached, and could result in a lot of unnecessary mongo queries. (getting slightly off-topic there)

I think an example of a job which never calls `done` for the first 2 tries shows the default behaviour of the lock more clearly, but there seems to be some strong opinions on which approach to take.

¯\\\_(ツ)_/¯

gianpaj:
Could this be merged? It took me a long time to find this PR, which code is pretty helpful.

Thanks

zentrapool:
I've had problem dealing abnormal behaviors with long running jobs. 
This code is very useful. hope this PR can be merge to help many Agenda users.

OmgImAlexis:
Is this able to be updated or should it be closed?

simison:
cc @koresar @leonardlin thoughts on this one or shall we close?

koresar:
I don't have an opinion. But few people said that it's a useful thing.
I'd rather have an example like that.

Can't be merged yet because CI...

santiq:
IMO this example is pretty useful, since it ilustrates how to use `job.touch()` with `lockLifetime`

leonardlin:
oh, it fails because of linting errors, hmm, we need to make a big linting fix PR

Otherwise I have nothing against the example.

simison:
Rebased and sorted lint errors.

simison (approved):

PR #797, opened 2019-04-03
.save and .remove return Promises and don't take callbacks

dandv:
See https://github.com/agenda/agenda/blob/master/lib/job/save.js for .save - it explicitly disallows passing a callback.

.remove should do the same, but it just doesn't call any callback - https://github.com/agenda/agenda/blob/master/lib/agenda/cancel.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -507,7 +507,7 @@ the database. See below to learn how to manually work with jobs.
 
 ```js
 const job = agenda.create('printAnalyticsReport', {userCount: 100});
-job.save(err => {
+await job.save(err => {
   console.log('Job successfully saved');
 });
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -507,7 +507,7 @@ the database. See below to learn how to manually work with jobs.
 
 ```js
 const job = agenda.create('printAnalyticsReport', {userCount: 100});
-job.save(err => {
+await job.save(err => {
```
wingsbob: That's a callback? 😲

README.md
```diff
@@ -709,27 +709,29 @@ job.run((err, job) => {
 });
 ```
 
-### save(callback)
+### save()
 
-Saves the `job.attrs` into the database.
+Saves the `job.attrs` into the database. Returns a Promise resolving to a Job instance, or throws on error.
```
wingsbob: Do we wanna say `rejects promise`?

PR #825, opened 2019-06-13
add support for limitless concurrency

freisenhauer:
In case you would like to manage you resource usage limits for example by setting limits in kubernetes, the concurrency limits are kind of a synthetic bottleneck. I have a usecase, where these limits are very annoying and setting them to limitless is way more elegant than setting a high value and keep this value in mind, in case something does not work some times.


MichielDeMey:
How would setting resource limits in Kubernetes help with this?

If you were to set the concurrency to an infinite number of jobs, Agenda might still crash due to out-of-memory issues caused by your resource limits, no?

PR #827, opened 2019-06-20
Solved: Job which are queued with high priority were not given priority 

MrSpark2591:
//registering jobs 
    const regJobToProcess = (job) => {
        let job = agenda.create(job.type, job);
        job.schedule('now');
        job.priority(notification.priority || 'low');
        job.save();
    };
    setTimeout(() => {
        regJobToProcess({ type: 'PROCESS_SMS_OTP', priority: 'low', data: 1 });
        regJobToProcess({ type: 'PROCESS_SMS_OTP', priority: 'high', data: 2 });
    }, 5000);
 
When next job is picked, I expected that it will be the job with high priority. But result was not what i expected and It was the one with low priority. So I have gone through some of the code and found out that sorting order was not right and it was taking nextRunTime 1st and than priority. But for priority queue it should be other way around.

Let me know what is the expected behaviour? If current behaviour is not what is expected please review and merge the PR.

### DIFF
--- a/lib/agenda/index.js
+++ b/lib/agenda/index.js
@@ -40,7 +40,7 @@ class Agenda extends EventEmitter {
     this._lockedJobs = [];
     this._jobQueue = [];
     this._defaultLockLifetime = config.defaultLockLifetime || 10 * 60 * 1000; // 10 minute default lockLifetime
-    this._sort = config.sort || {nextRunAt: 1, priority: -1};
+    this._sort = config.sort || {priority: -1, nextRunAt: 1};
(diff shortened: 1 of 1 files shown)


Uzlopak:
seems legit

simllll:
Beware this could be a breaking change.
1.) Behavior changes
2.) Database query changes (other sorting/order means different index is required

MrSpark2591:
> Beware this could be a breaking change.
> 1.) Behavior changes
> 2.) Database query changes (other sorting/order means different index is required

I guess index `findAndLockNextJobIndex` will be used no need to change it. And as for behavior change, I guess the proposed behavior is the correct one. Let me know if I am missing something.

PR #836, opened 2019-07-23
Remove unsupported callback params

lqmanh:
This PR removes all unsupported "cb" params in `agenda.every()`, `agenda.now()` and `agenda.schedule()`.
There are #752 and #831 but they are incomplete. This should be an alternative to both of them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,7 +387,7 @@ agenda.define('say hello', job => {
 
 ## Creating Jobs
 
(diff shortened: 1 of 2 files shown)


lqmanh:
There was something wrong with the job lock, which made the Travis CI build failed . I don't know why, but I'm sure it has nothing to do with this PR.

simison:
Done elsewhere already, thanks!

PR #852, opened 2019-09-09
Fixes for failing test cases

dmbarreiro:
These modifications try to fix an issue with failing test cases which are preventing contributions to make it to the repo (when these contributinos have nothing to do with the tests themselves).

The changes are:
* new queuing mechanism since the previous didn't guarantee priority when executing jobs scheduled the same datetime (agenda uses setTiemout and sometimes the job with lowest priority was executed, though not processed, first). Test case involved: job concurrency -> 'should run jobs as first in first out (FIFO) with respect to priority'

* On repeatEvery -> 'sets the nextRunAt property with skipImmediate' test case we expect precise timing in miliseconds and that's not always the with the processing we do to set repeatEvery, sometimes we schedule the job one milisecond after the expected time, I find this acceptable and changed the test case to expect something in the range of 3 minutes sharp and 3 minutes + 2 miliseconds for nextRunAt.

* start/stop -> 'clears locks on stop' fails on many PRs not being able to find 'longRunningJob' on db. I provided the param job on define callback since not having params has caused trouble before and increased jobTimeout when we execute the tests on TravisCI (actually I only set the TRAVIS env var on .travis.yml, the tests were already prepared to increase this timeout when that env var is present). Hopefully it'll help, never suffered the issue when testing locally, I've only seen it on TravisCI.

### DIFF
--- a/lib/agenda/job-processing-queue.js
+++ b/lib/agenda/job-processing-queue.js
@@ -1,0 +1,82 @@
+/**
+ * @class
+ * @param {Object} args - Job Options
+ * @property {Object} agenda - The Agenda instance
+ * @property {Object} attrs
+ */
+class JobProcessingQueue {
+  constructor() {
+    this._queue = [];
+  }
+
+  get length() {
+    return this._queue.length;
+  }
+}
+
+/**
+ * Pops and returns last queue element (next job to be processed) without checking concurrency.
+ * @returns {Job} Next Job to be processed
+ */
+JobProcessingQueue.prototype.pop = function() {
+  return this._queue.pop();
+};
+
+/**
+ * Inserts job in first queue position
+ * @param {Job} job job to add to queue
+ * @returns {undefined}
+ */
+JobProcessingQueue.prototype.push = function(job) {
+  this._queue.push(job);
+};
+
+/**
+ * Inserts job in queue where it will be order from left to right in decreasing
+ * order of nextRunAt and priority (in case of same nextRunAt), if all values
+ * are even the first jobs to be introduced will have priority
+ * @param {Job} job job to add to queue
+ * @returns {undefined}
+ */
+JobProcessingQueue.prototype.insert = function(job) {
+  let matchIndex;
+  this._queue.some((element, index) => {
+    if (element.attrs.nextRunAt.getTime() <= job.attrs.nextRunAt.getTime()) {
+      if (element.attrs.nextRunAt.getTime() === job.attrs.nextRunAt.getTime()) {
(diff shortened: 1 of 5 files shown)


lib/agenda/job-processing-queue.js
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Inserts job in queue where it will be order from left to right in decreasing
+ * order of nextRunAt and priority (in case of same nextRunAt), if all values
+ * are even the first jobs to be introduced will have priority
+ * @param {Job} job job to add to queue
+ * @returns {undefined}
+ */
+JobProcessingQueue.prototype.insert = function(job) {
+  let matchIndex;
+  this._queue.some((element, index) => {
```
koresar: Since node.js 4 can be simplified to:
```js
const matchIndex = this._queue.findIndex(element => {
    if (element.attrs.nextRunAt.getTime() <= job.attrs.nextRunAt.getTime()) {
      if (element.attrs.nextRunAt.getTime() === job.attrs.nextRunAt.getTime()) {
        if (element.attrs.priority >= job.attrs.priority) {
          return true;
        }
      } else {
        return true;
      }
    }
    return false;
  });
```

dmbarreiro: Thank you! Your changes are in now. One of the test cases I was trying to fix failed now for Node 10 so I'll have a look again and try to fix it properly.

test/job.js
```diff
@@ -615,22 +615,21 @@ describe('Job', () => {
...
-      await delay(jobTimeout);
-      await agenda.stop();
+      try {
+        agenda.define('longRunningJob', job => { // eslint-disable-line no-unused-vars
+          // Job never finishes
+        });
+        agenda.every('10 seconds', 'longRunningJob');
+        agenda.processEvery('1 second');
 
-      agenda._collection.findOne({name: 'longRunningJob'}, (err, job) => {
```
dmbarreiro: I suspect there was an await missing here, that caused the afterEach mocha section of the test to be executed clearing the jobs which eventually made the findOne return undefined, no job found.

test/job.js
```diff
@@ -613,22 +615,21 @@ describe('Job', () => {
...
-        if (err) {
-          throw err;
-        }
+        await agenda.start();
+        await delay(jobTimeout);
+        await agenda.stop();
+        const job = await agenda._collection.findOne({name: 'longRunningJob'});
         expect(job.lockedAt).to.be(null);
-      });
+      } catch (err) {
```
koresar: Since you fixed it, the try+catch+throw can be removed now. It's the default behaviour anyway.

Issue #859, opened 2019-09-19
skipImmediate causes repeat job scheduling to fail


AshlinDuncan:
According to #758, a repeating event that begins repeating on a particular date can be set like this:

    job.schedule('in 3 days').repeatEvery('20 minutes');

When three days pass, the job will begin running every twenty minutes. This is expected behavior.

However, it has a problem. When three days is reached, it will run immediately, rather than first waiting 20 minutes.

If the event was starting now instead of in three days, this would be fixed by the following code: 

    job.schedule('in 3 days').repeatEvery('20 minutes', {
        skipImmediate: true
    });

However, including `skipImmediate` seems to completely negate the `.schedule('in 3 days')` command, and the event begins in twenty seconds from now instead. It seems the implementation of `skipImmediate` does not consider the initial start time -- only the repeat time.

In my project, I am fixing this by creating a job *that creates a job* to run the repeat interval. However, this is clunky.

PR #860, opened 2019-09-19
Fix for skipImmediate resetting nextRunAt to current date

AshlinDuncan:
Previously, as in issue #859, skipImmediate would always reset nextRunAt to the current date plus the increment, rather than maintaining the existing nextRunAt. This pull request solves this.


koresar:
Hello @AshlinDuncan 
Looks like the CI build has failed. Could you please look at that?

lib/job/repeat-every.js
```diff
@@ -13,7 +13,8 @@ module.exports = function(interval, options) {
   this.attrs.repeatInterval = interval;
   this.attrs.repeatTimezone = options.timezone ? options.timezone : null;
   if (options.skipImmediate) {
-    this.attrs.lastRunAt = new Date();
+    // Set the lastRunAt time to the nextRunAt so that the new nextRunAt will be computed in reference to the current value.
+    this.attrs.lastRunAt = this.attrs.nextRunAt;
```
koresar: I believe this change belongs to the `computeNextRunAt()` function. It's better be implemented there not in this file.

AshlinDuncan: Thanks for the review! Could you elaborate? I'm not sure what you mean.
The `skipImmediate` option has always been calculated in `repeatEvery`, and `this.computeNextRunAt()` called afterwards to "skip" the immediate call, in order to avoid modifying `computeNextRunAt`. The change I made simply allows it to "remember" the previous `nextRunAt` value in the case of a chained call, like `job.schedule('in 3 days').repeatEvery('20 minutes')` (previously, the "in 3 days" part would be wiped by hard setting the lastRunAt to now). My change keeps the skipImmediate implementation within repeatEvery, as it was before.
Is your change request to move the `skipImmediate` implementation outside of repeatEvery and into computeNextRunAt entirely? If so, I think the most reasonable approach would be to keep track of a "skipNext" property in `attrs`. If true, `computeNextRunAt` would calculate an extra interval forward, and set skipNext to false. Would you agree?
If you wanted to keep the `repeatEvery` implementation, and just change `computeNextRunAt`, I'm not sure what you mean since the part where `lastRunAt` is set to now is the buggy part.

koresar: Ok. I hear your point. You are right.
Let's make this line like this for safety reasons:

```js
this.attrs.lastRunAt = this.attrs.nextRunAt || new Date();
```

(Btw, unit tests are still failing.)

AshlinDuncan: Committed! Thank you!
In terms of unit tests, the "sets the nextRunAt property with skipImmediate" test is the one that is failing. I did edit this test. Before, it was calling `new Date()` with the `repeatEvery` call instead of the `new Job()` call. Due to my change, it uses the date defined in `new Job()` (as it should), so to check equality with that one, they must be declared at (about) the same time. 
I would guess that the Travis CI build is running on the old tests, and not my edited version.
Alternatively, it could be using my edited version, but there's a significant delay between `new Job()` and the next line, `new Date()`. That said, on my end, those tests pass just fine. Do you know how to get Travis to use my new tests, or how to figure out whether it is already?

It also looks like there's an additional failure on node versions 8 and 10 in the afterEach hook, however, that has nothing to do with this pull request and exists in the master branch and in the majority of recent pull requests.

koresar: You would need to synchronise your branch with the latest `master`. You'd need to merge agenda's latest `master` to your branch.

Or start from scratch. There are not that many changes.

koresar: The unit test would be fixed if you add that ` || new Date()` I suggested.

I don't know how to fix node.js 8 and 10 CI build. If you could do it - everyone would be happy! Including other contributors.

koresar: I believe that instead of creating a new unit test you'd better amend this test: `'sets the nextRunAt property with skipImmediate'` to your needs.

AshlinDuncan: Thank you!
I went ahead and merged, everything should be passing now.

MichielDeMey:
@koresar Are there any outstanding issues that would prevent this from being merged upstream? 🙂

koresar:
@MichielDeMey yeah. The unit test needs fixing. Otherwise it will be randomly failing.

MichielDeMey:
@koresar They look ok to me, any specific flakiness behavior we should watch out for?

koresar:
Soory for screenshotting.

image

koresar:
I know why no one saw my review comments. I needed to click "Submit Review". :(

I'm so sorry!

test/job.js
```diff
@@ -99,11 +99,17 @@ describe('Job', () => {
...
     });
     it('sets the nextRunAt property with skipImmediate', () => {
+      const job2 = new Job();
       const now = new Date();
-      job.repeatEvery('3 minutes', {skipImmediate: true});
-      const lowerBound = now.valueOf() + 180000;
-      const upperBound = now.valueOf() + 180000 + 2;
-      expect(job.attrs.nextRunAt).to.be.within(lowerBound, upperBound); // Inclusive
+      job2.repeatEvery('3 minutes', {skipImmediate: true});
+      expect(job2.attrs.nextRunAt).to.be(now.valueOf() + 180000); // Inclusive
```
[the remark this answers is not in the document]
AshlinDuncan: You're right! Apologies about that. I am fixing it in the next commit.

test/job.js
```diff
@@ -99,11 +99,17 @@ describe('Job', () => {
...
+    });
+    it('repeats from the existing nextRunAt property with skipImmediate', () => {
+      const job2 = new Job();
+      const futureDate = (new Date('3000-01-01T00:00:00')).valueOf();
+      job2.attrs.nextRunAt = futureDate;
```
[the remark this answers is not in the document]
AshlinDuncan: I'll add another test to cover that, if that works? To check both the specific implementation bug and the specification bug. 
(Note: Current build not passing; I'll fix it momentarily.)

PR #867, opened 2019-10-11
Update README.md

lautarobock:
Hello, I've created a project to help to use agenda with TypeScript and [Inversify](https://github.com/inversify/InversifyJS), it is pretty simple but could help to define workers with a class and organize your project. 
There are a lot of work to do, but it is already in use and it is open source and open for collaboration.
PS: Not sure if is the way to share my project but I can fin another way.
PS#2: It is already published in the [Inversify docs](https://github.com/inversify/InversifyJS/blob/master/wiki/ecosystem.md#extensions)


simison:
Neat!

PR #874, opened 2019-11-04
Update documentation for `Agenda#now`

mmiller42:
[`Agenda#now`](https://github.com/agenda/agenda/blob/master/lib/agenda/now.js) doesn't appear to return the job (for chaining). Unlike other scheduling methods, `now` immediately saves the job.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,7 +473,7 @@ Schedules a job to run `name` once immediately.
 `data` is an optional argument that will be passed to the processing function
 under `job.attrs.data`.
 
-Returns the `job`.
+Saves the job instance and returns a promise which resolves with the `job`.
 
 ```js
 agenda.now('do the hokey pokey');
@@ -978,11 +978,17 @@ let app = express(),
 
 app.post('/users', (req, res, next) => {
   const user = new User(req.body);
-  user.save(err => {
+  user.save(async err => {
     if (err) {
       return next(err);
     }
-    agenda.now('registration email', {userId: user.primary()});
+    
+    try {
+      await agenda.now('registration email', {userId: user.primary()});
+    } catch (err) {
+      return next(err);
+    }
+    
     res.send(201, user.toJson());
   });
 });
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -473,7 +473,7 @@ Schedules a job to run `name` once immediately.
 `data` is an optional argument that will be passed to the processing function
 under `job.attrs.data`.
 
-Returns the `job`.
+Saves the job instance and returns a promise which resolves with the `job`.
```
sampathBlam: @mmiller42 : Can we change the text to `Returns a promise which resolves with the created job or rejects with an error`?

README.md
```diff
@@ -473,7 +473,7 @@ Schedules a job to run `name` once immediately.
...
 
 ```js
 agenda.now('do the hokey pokey');
```
sampathBlam: @mmiller42 : Pls update this line in the documentation as well. We need to add await to this line. 
`const job = await agenda.now('do the hokey pokey');`

README.md
```diff
@@ -656,7 +656,7 @@ the following:
 - `insertOnly`: `boolean` will prevent any properties from persisting if the job already exists. Defaults to false.
 
 ```js
-job.unique({'data.type': 'active', 'data.userId': '123', nextRunAt(date)});
+job.unique({'data.type': 'active', 'data.userId': '123'});
```
mmiller42: Not sure if this could be replaced with something else, but the current syntax was invalid and I'm not sure what purpose a unique constraint on `nextRunAt` would serve.

sampathBlam: @mmiller42 : I believe you are right. But should we have it as a separate PR?

mmiller42: removed

README.md
```diff
@@ -978,11 +978,17 @@ let app = express(),
 
 app.post('/users', (req, res, next) => {
   const user = new User(req.body);
-  user.save(err => {
+  user.save(async err => {
```
sampathBlam: @mmiller42 : Can we replace the callback based example of `save` method with async/await? The middleware for '/users' would then become an async method. And we would have a single try/catch to handle errors from both `user.save` and `agenda.now`. Does that make sense?

PR #876, opened 2019-11-20
fix: Add try/catch block to agenda#now method

sampathBlam:
- Enclose agenda#now with try catch as it is an async method
- throw error in case of an error in creating/saving the job.
- Supplements documentation changes mentioned in #874


simison:
Thank you!

PR #887, opened 2019-12-03
Fix periodic node.js process unhandledRejection

koresar:
This PR is a clone of @Scorpil work in #886 , I just added a unit test.
My servers are currently crashing because of this bug. Please, process the PR asap.
Below is the copy-pasta from #886

======

Job processing is called from `processJobResult`, so calling it directly in `runOrRetry` is unnecessary. If executed job is short-lived, it leads to two `jobProcessing` calls occuring at almost the same time, and if there is another job it a queue waiting to be scheduled it will be scheduled twice (see the call to `runOrRetry` with `setTimeout` on line 217). On the second run the queue will be empty, which leads to `UnhandledPromiseRejectionWarning`. Error is thrown on line 228, because `jobQueue.pop()` returns undefined when queue is empty.

Script to help reproduce the issue:
```javascript
const Agenda = require("agenda");
const agenda = new Agenda({ db: { address: "<mongo_url>", collection: "ag_test" }});

agenda.define("j1", (job, done) => {
    console.log(Date.now(), 'j1');
    done();
});

agenda.define("j2", (job, done) => {
    console.log(Date.now(), 'j2');
    done();
});


(async function() {
    await agenda.start();
    await agenda.every("5 seconds", "j1");
    await agenda.every("10 seconds", "j2");
})();
```

Example output:
```
node test.js
(node:17695) DeprecationWarning: current Server Discovery and Monitoring engine is deprecated, and will be removed in a future version. To use the new Server Discover and Monitoring engine, pass option { useUnifiedTopology: true } to the MongoClient constructor.
1575304118915 'j2'
1575304119055 'j1'
1575304123786 'j1'
(node:17695) UnhandledPromiseRejectionWarning: TypeError: Cannot read property 'attrs' of undefined
    at Timeout.runOrRetry [as _onTimeout] (/Users/*/Code/agenda/lib/utils/process-jobs.js:228:47)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
(node:17695) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:17695) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

### DIFF
--- a/test/agenda.js
+++ b/test/agenda.js
@@ -26,7 +26,7 @@ const jobTimeout = process.env.TRAVIS ? 2500 : 500;
 const jobType = 'do work';
 const jobProcessor = () => {};
 
-describe('Agenda', () => {
+describe('Agenda', function() {
   beforeEach(() => {
     // @TODO: this lint issue should be looked into: https://eslint.org/docs/rules/no-async-promise-executor
     // eslint-disable-next-line no-async-promise-executor
@@ -566,6 +566,40 @@ describe('Agenda', () => {
       expect(job1.attrs.data).to.be(3);
       expect(job2.attrs.data).to.be(2);
       expect(job3.attrs.data).to.be(1);
+    });
+  });
+
+  describe("process jobs", function() {
+    it("should not process twice", async function() {
+      this.timeout(10000);
+
+      let unhandledRejections = [];
+      const rejectionsHandler = error => unhandledRejections.push(error);
+      process.on('unhandledRejection', rejectionsHandler);
+
+      let j1processes = 0;
+      jobs.define("j1", (job, done) => {
+        j1processes += 1;
+        done();
+      });
+
+      let j2processes = 0;
+      jobs.define("j2", (job, done) => {
+        j2processes += 1;
+        done();
+      });
+
+      await jobs.start();
+      await jobs.every("5 seconds", "j1");
+      await jobs.every("10 seconds", "j2");
+
+      await delay(6000);
+      process.removeListener('unhandledRejection', rejectionsHandler);
+
+      expect(j1processes).to.equal(2);
+      expect(j2processes).to.equal(1);
+
+      expect(unhandledRejections).to.have.length(0);
     });
   });
 });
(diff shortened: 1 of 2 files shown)


Scorpil:
Thanks for the unittest 👍 I'm closing the old PR.

YkmLo:
hi, when will this change be released? i have the same problem

koresar:
@simison please help.

test/agenda.js
```diff
@@ -26,7 +26,7 @@ const jobTimeout = process.env.TRAVIS ? 2500 : 500;
 const jobType = 'do work';
 const jobProcessor = () => {};
 
-describe('Agenda', () => {
+describe('Agenda', function() { // eslint-disable-line prefer-arrow-callback
```
simison: Is there a reason not to use arrow function here?

test/agenda.js
```diff
@@ -568,4 +568,41 @@ describe('Agenda', () => {
...
+      jobs.define('j2', (job, done) => {
+        j2processes += 1;
+        done();
+      });
+
+      await jobs.start();
+      await jobs.every('5 seconds', 'j1');
+      await jobs.every('10 seconds', 'j2');
+
+      await delay(6000);
```
simison: Why this long delay? Could the test be structured without it (or with shorter?)

Edit: oh I saw your code-comment above. Still would love if we'd have something else than timeout to test.

simison:
Merging as-is, we can loop back to questions in separate pr.

test/agenda.js
koresar: To make the bottom this.timeout() go work

test/agenda.js
koresar: I tried shortening all the delays x10. For some reason the bug didn't reproduce. Not sure why.
It's likely can be shortened. I just didn't have much time to investigate. Sorry.

koresar:
Can we expect the bug fix release sometime before the NY? 😁

simison:
Yep, I'll roll it today soonish.

PR #894, opened 2019-12-15
Pause and Resume implementation wrapper using already exisiting functions

mur2zab:
Couldn't find pause and resume functionality in the existing library. Tried to help using the existing functions. Comments are appreciated. **First contribution to Open Source**.

### DIFF
--- a/lib/job/pause.js
+++ b/lib/job/pause.js
@@ -1,0 +1,14 @@
+'use strict';
+const noCallback = require('../no-callback');
+
+/**
+ * Pause a job into the MongoDB
+ * @name Job#
+ * @function
+ * @returns {Promise} instance of Job resolved after job is paused or errors
+ */
+module.exports = function() {
+  noCallback(arguments);
+  this.attrs.disabled = true;
+  return this.agenda.saveJob(this);
+};
--- a/lib/job/resume.js
+++ b/lib/job/resume.js
@@ -1,0 +1,18 @@
+'use strict';
+const date = require('date.js');
+
+/**
+ * Resumes a job into the MongoDB
+ * @name Job#
+ * @function
+ * @param when time for the new nextRunAt field 
+ * @returns {Promise} instance of Job that is resumed
+ */
+module.exports = function(when) {
+
+  const d = new Date(when);
+  this.attrs.disabled = false;
+  this.attrs.nextRunAt = Number.isNaN(d.getTime()) ? date(when) : d;
+  return this;
+
+};
--- a/lib/job/index.js
+++ b/lib/job/index.js
@@ -68,5 +68,7 @@ Job.prototype.isRunning = require('./is-running');
 Job.prototype.save = require('./save');
 Job.prototype.remove = require('./remove');
 Job.prototype.touch = require('./touch');
+Job.prototype.pause = require('./pause');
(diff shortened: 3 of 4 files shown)


AhmadIbrahiim:
It seems to be perfect if we can add **pause** and **resume**  to AGENDA!

simison:
`pause('1 day')` could be an interesting feature as well (using [`human-interval`](https://github.com/agenda/human-interval). No need in this PR, just noting for future.

simison:
`Should pause the job`  test seems to be failing at the moment. @mur2zab  could you have a look? Thanks!

lib/job/pause.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+/**
+ * Pause a job into the MongoDB
+ * @name Job#
```
simison: Could you change this to `@name @name Job#pause`?

lib/job/resume.js
```diff
@@ -0,0 +1,16 @@
+'use strict';
+const date = require('date.js');
+
+/**
+ * Resumes a job into the MongoDB
+ * @name Job#
```
simison: Could you change this to `@name @name Job#resume`?

lib/job/resume.js
```diff
@@ -0,0 +1,16 @@
...
+ * @function
+ * @param when time for the new nextRunAt field
+ * @returns {Promise} instance of Job that is resumed
+ */
+module.exports = function(when) {
```
simison: How would we resume jobs with an interval set? Should we use `computerInterval`?

https://github.com/agenda/agenda/blob/b71d3eaecc9b6eb7af976a165b8fb27dcbebac24/lib/job/compute-next-run-at.js#L34

simison:
I'd like to see tests that cover different types of jobs: requrring ones, one-off ones etc. What do you think?

shirohige:
> `Should pause the job` test seems to be failing at the moment. @mur2zab could you have a look? Thanks!
> 
> Could you also add this to documentation (in `README.md`)

Sure will look into this. Thanks for reviewing @simison

mur2zab:
> `Should pause the job` test seems to be failing at the moment. @mur2zab could you have a look? Thanks!
> 
> Could you also add this to documentation (in `README.md`)

Sure will look into the comments. Thanks for reviewing @simison

pdfowler:
Proposing we close this as pause/resume implementation is near identical to the existing enable/disable functionality.

[From PR in the `pause.js` file](https://github.com/agenda/agenda/pull/894/files#diff-7fd9318bca08181664bc70e8b8f35bbacaea20bd912a381bf06301e77d26aa1aR8-R11)
```
// Pause
module.exports = function() {
   this.attrs.disabled = true;
   return this.save();
 };
```

[From existing `disable` file](https://github.com/agenda/agenda/blob/7e78b71b79ae6c6d874e9180954b8b84248e9848/lib/job/disable.ts#L8-L11)
```
// Disable
export const disable = function (this: Job): Job {
  this.attrs.disabled = true;
  return this;
};
```

From a DX perspective, I often got tripped up with which job methods called `save` internally and which were chainable. I have a strong preference for chainable job methods, and leaving it to the user to call save explicitly.

PR #904, opened 2019-12-27
Fix job queue issues

Seandor:
This PR is related to https://github.com/agenda/agenda/pull/887
We cannot solve one problem and create another.

>>> "Job processing is called from processJobResult, so calling it directly in runOrRetry is unnecessary."

It's necessary because without `jobProcessing` after each run of a job, jobs in the jobQueue might not be able to run on time. `jobProcessing` isn't always called within `processJobResult`, check the first line of the function. 

My fix to the issue "unhandledRejection" is to revert the code to the old version of Agenda. We should not pop job in `runOrRertry` function, because it makes `returnNextConcurrencyFreeJob` pointless, the "conconrrent free" jobs will never be used. The reason that `runOrRetry` is an internal function in `jobProcessing` is because it wants to use the job in the lexical scope.

Will add examples later.

### DIFF
--- a/lib/agenda/job-processing-queue.js
+++ b/lib/agenda/job-processing-queue.js
@@ -61,7 +61,7 @@ JobProcessingQueue.prototype.insert = function(job) {
 };
 
 /**
- * Returns (does not pop, element remains in queue) first element (always from the right)
+ * Returns first element (always from the right)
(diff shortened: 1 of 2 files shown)


Seandor:
The job queue is still not quite right. closing this.

PR #925, opened 2020-02-13
Drop Node.js 8 support

simison:
Drops Node.js 8 support. I mainly just wanted to update dependencies to the latest and at least `xo` (dev-dependency) isn't compatible with Node.js 8 anymore. 

The maintenance period for 8 has finished anyway so I think this is an ok move: https://nodejs.org/en/about/releases/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,6 @@ language: node_js
 services: mongodb
 sudo: required
 node_js:
-  - 8
   - 10
   - 12
 addons:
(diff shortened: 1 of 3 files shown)


MichielDeMey:
Makes sense to drop Node 8 since its maintenance period has expired. 👍

simison:
Thanks! @MichielDeMey you should join our volunteer Slack! :-) Link at the readme.

PR #948, opened 2020-02-24
DeprecationWarnings: autoReconnect, reconnectTries, reconnectInterval

ekegodigital:
I'm using the Docker Image mongo:3.6 and switched the "useUnifiedTopology" parameter to true as advised. If I do so I get the following deprecation warnings:

(node:51) DeprecationWarning: The option autoReconnect is incompatible with the unified topology, please read more by visiting http://bit.ly/2D8WfT6
(node:51) DeprecationWarning: The option reconnectTries is incompatible with the unified topology, please read more by visiting http://bit.ly/2D8WfT6
(node:51) DeprecationWarning: The option reconnectInterval is incompatible with the unified topology, please read more by visiting http://bit.ly/2D8WfT6

The fix checks if "useUnifiedTopology" is set and then doesn't use the deprecated options. If "useUnifiedTopology" isn't used, the options stay the same.

### DIFF
--- a/lib/agenda/database.js
+++ b/lib/agenda/database.js
@@ -26,7 +26,7 @@ module.exports = function(url, collection, options, cb) {
   }
 
   let reconnectOptions = {autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery};
-  if (options && typeof options.useUnifiedTopology !== 'undefined' && options.useUnifiedTopology === true) {
+  if (options && options.useUnifiedTopology && options.useUnifiedTopology === true) {
     reconnectOptions = {};
   }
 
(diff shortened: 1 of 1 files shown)


MichielDeMey:
Thanks for this PR, before we can merge you need to make sure the linting/test pass on out CI. 😉

lib/agenda/database.js
```diff
@@ -25,8 +25,13 @@ module.exports = function(url, collection, options, cb) {
     url = 'mongodb://' + url;
   }
 
+  let reconnectOptions = {autoReconnect: true, reconnectTries: Number.MAX_SAFE_INTEGER, reconnectInterval: this._processEvery};
+  if (options && typeof options.useUnifiedTopology !== 'undefined' && options.useUnifiedTopology === true) {
```
MichielDeMey: I think we can improve this to a falsy check to also accommodate `null`, `undefined` and `false`.

```suggestion
  if (options && options.useUnifiedTopology && options.useUnifiedTopology === true) {
```

What do you think?

PR #1096, opened 2020-06-26
Updated the History.md

maabiddevra:
MongoDB's Unified Topology feature is rolled out in version `3.1.0` but in `History.md` it's mention in `3.0.0`, fixed it.

### DIFF
--- a/History.md
+++ b/History.md
@@ -8,6 +8,7 @@ Next
 
 _Stay safe!_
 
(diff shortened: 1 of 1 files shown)


simison:
Sure? 

Commit log tells it went to 3.0.0 like History.md says:

image

Unless I'm missing something?

PR #1078, opened 2020-05-13
add mongodb-memory-server

Uzlopak:
adding mongodb-memory-server

some tests fail, but i dont know how to fix them.

PR #1108, opened 2020-08-13
Add mongodb memory server

pdfowler:
I have a fair amount of experience with `mongo-memory-server` and after seeing the difficulties with PR #1078 I thought I'd see if I could get things working. 

## Highlights

1. By default tests will be executed against an in-memory instance of MongoDB
2. Tests can be executed against a hosted MongoDB instance by setting the `USE_MONGODB` env variable to a truthy value (yes|true|1).
3. Travis and Docker configurations will not use the new in-memory db instance.

### Misc
* Some fragile tests were "hardened"
* Logging adjustments
* Linting configuration specific to the `/test` directory

## Todo

- [x] Look into auto formatting abnormalities (ie: `test/retry.js` appears to have tabs in the diff)
- [ ] Set a fixed MongoDB version for `mongo-memory-server`


Uzlopak:
I saw your PR already a week ago. Actually I ported for our needs agendajs to typescript and also use mongoose instead of direct mongo driver. So for us all is up and running and "fully" tested. Somehow agendajs is janky and I am kind of feeling that it is kind of unstable. So I am happy if there are better tests/code so that I can port it to our implementation.

koresar:
As a hardcore node developer I'd like to clap to this PR. Brilliant!

Changes look minimalistic clean and thoroughly thought through.

I've pulled the branch, run the tests on my machine, - all good!

Let's merge it.

Unrelated: I tried replacing all `delay(50)` with `delay(5)` and the testing time dropped from 45 to 24 seconds.

koresar (dismissed):

simison:
@Uzlopak @pdfowler ah bummer, our typescript rewrite stepped on this PR just before merging.

Would you be so kind and follow-up on renamed `.ts` files and we can then merge this? Sorry for inconvenience and thanks for your work thus far!

simison:
>    Some fragile tests were "hardened"
>    Logging adjustments
>    Linting configuration specific to the /test directory

These might better be in their own PRs so that it's easier to review/rebase/merge.

test/job.js
```diff
@@ -10,66 +10,45 @@ const delay = require("delay");
...
 let mongoClient = null;
 
-const clearJobs = () => {
-  return mongoDb.collection("agendaJobs").deleteMany({});
-};
-
-// Slow timeouts for CI
-const jobTimeout = 500;
+// Slow timeouts for Travis
+const jobTimeout = process.env.TRAVIS ? 2500 : 500;
```
pdfowler, line 22: Seems that travis got dropped since my original work on this ... tagging line for removal

test/job.js
```diff
@@ -10,66 +10,45 @@ const delay = require("delay");
...
-        if (error) {
-          done(error);
-        }
+    });
 
-        try {
-          const client = await MongoClient.connect(mongoCfg, {
-            useNewUrlParser: true,
-            useUnifiedTopology: true,
-          });
```
pdfowler, line 51: I'm not clear on the order of changes WRT the mongo connect options. Were these part of new tests/functionality or are they safe to ignore?

pdfowler:
@simison I finally got a chance to dust off the code and get upstream/master merged in. After further review, I'm left with the following questions/issues/outstanding tasks:

[] Extracting out the non-core code from this PR (partially resolved w/ elimination of `xo`)
[] Determine whether the [mongo config options](https://github.com/agenda/agenda/pull/1108#pullrequestreview-692525434) are core to the tests (currently this PR removes them)
[] Decide on behavior of failing test: [job methods:unique:...:should modify one job when unique matches](https://github.com/agenda/agenda/blob/c29d2141a314429f2600bac184a26167ae1fcf8f/test/agenda.js#L416)

test/agenda.js
```diff
@@ -414,7 +409,10 @@ describe("Agenda", () => {
             })
             .schedule("now")
             .save();
+            debug('Job 2 scheduled for ' + job2.attrs.nextRunAt.toISOString())
 
+          // TODO: This test is currently failing, and I have to ask whether the test describes the expected behavior
+          // since the 'insertOnly' flag is not set on the `unique` call
           expect(job1.attrs.nextRunAt.toISOString()).not.to.equal(
```
pdfowler, line 416: This test is failing on both this branch and master

test/job.js
koresar: Yes, please remove travis. Instead you might want to use the standard `CI` env var.

```
if (process.env.CI)
```

test/job.js
koresar: They are useful for old mongodb versions. I think prior to 3.6. But we do not support older mongodb any more. So, it's okay to drop these options.

koresar (changes requested):
Just replace `process.env.TRAVIS` with the more universal `process.env.CI` and we are good to go!

test/job.js
pdfowler: @koresar Github PR diffs can be a bit strange at times: that line was still showing in the PR, but had been removed when I pulled it locally. I've pushed a change to restore it to the flat `const jobTimeout = 50;`

pdfowler:
@koresar Just pushed the change to restore the `jobTimeout` variable to use the `process.env.CI` ([test/job.js](https://github.com/agenda/agenda/pull/1108/files?file-filters%5B%5D=.js&file-filters%5B%5D=.lock&file-filters%5B%5D=.yml#diff-3d864fcfcae7cc47fa275a1fac955f5a2e5f44538c11c65ea1342fbdd1717ed1R20)).

The only remaining question is WRT the failing "should modify one job when unique matches" test in `test/agenda.js`. I've [added a comment](https://github.com/agenda/agenda/pull/1108/files?file-filters%5B%5D=.js&file-filters%5B%5D=.lock&file-filters%5B%5D=.yml#diff-c5a2911b4cb8dd6e1674860f862942dbd34f7a91776c2e4b7de43728e5b26da2R415-R416) there pointing to a possible resolution, but wasn't sure it should be addressed as part of this PR or not.

test/agenda.js
```diff
@@ -416,7 +410,10 @@ describe("Agenda", () => {
             })
             .schedule("now")
             .save();
+            debug('Job 2 scheduled for ' + job2.attrs.nextRunAt.toISOString())
 
+          // TODO: This test is currently failing, and I have to ask whether the test describes the expected behavior
```
koresar, line 415: Please remove this comment. It's not relevant any more. The unit test got broken with a different PR. The test is passing now. And the test is correct. It must make sure that the `nextRunAt` changed. Later on it checks that there is only one job in the DB.

koresar (changes requested):

koresar:
@pdfowler I've just sent you a request to become one of the maintainers. You can now create branches directly within Agenda and all its repos. I hope you'll find it useful.
Also, you can now close tickets and merge PRs.

harisvsulaiman:
@pdfowler Welcome to the team. I hope we can transform agenda into the best job scheduler with good persistence gaurentees.

koresar (approved):

test/find-and-lock-next-job.js
```diff
@@ -24,42 +22,32 @@ const jobType = "do work";
...
-            agenda.define(jobType, jobProcessor);
-
-            done();
-          }
-        );
-      }
-    );
+      });
+  
+      await jobs._ready;
```
harisvsulaiman, line 37: There is no jobs here. Did you mean agenda? Related to https://github.com/agenda/agenda/issues/1369.

koresar: That's an after merge bug. Could you please fix it in `master` @harisvsulaiman ?

PR #1109, opened 2020-08-20
Add top level disable and enable

pdfowler:
## Use Case
We have a job which sends out invites in a staged manner over a period of time: "sendInvites". Once enough people have responded to the invite, we don't want to send out any more invites. If someone later cancels, we want to continue sending out invites in a tiered manner.

## Workaround
The same effect can be accomplished by querying any outstanding "sendInvites" jobs, and iterating over them and calling `job.disable()`, and doing the same thing to enable with `job.enable()`. 

Alternatively the `this._collection.updateMany` based implementation can be patched to a local instance via prototype injection.

---
## Proposed Implementation
Following the example of `agenda.cancel()` and its relationship with `job.remove()`, this PR adds the following two methods:

> ### disable(mongodb-native query)
> 
> Disables any jobs matching the passed mongodb-native query, preventing any matching jobs from being run by the Job Processor.
> 
> ```js
> const numDisabled = await agenda.disable({name: 'sendInvites'});
> ```
> 
> Similar to `agenda.cancel()`, this functionality can be acheived with a combination of `agenda.jobs()` and `job.disable()`
> 
> ### enable(mongodb-native query)
> 
> Enables any jobs matching the passed mongodb-native query, allowing any matching jobs to be run by the Job Processor.
> 
> ```js
> const numEnabled = await agenda.enable({name: 'sendInvites'});
> ```
> 
> Similar to `agenda.cancel()`, this functionality can be acheived with a combination of `agenda.jobs()` and `job.enable()`

---

## PR Checks:

- [x] Add Test Coverage
- [x] Add documentation to README.md
- [ ] Update HTML documentation


lib/agenda/enable.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+const debug = require('debug')('agenda:enable');
+
+/**
+ * Enables any jobs matching the passed MongoDB query, and removes them from the database.
```
koresar, line 5: "and removes them from the database" ?
I don't see that happening in the code. :) Perhaps, the comment needs an update?

lib/agenda/enable.js
```diff
@@ -0,0 +1,22 @@
...
+ * @name Agenda#enable
+ * @function
+ * @param {Object} query MongoDB query to use when enabling
+ * @caller client code, Agenda.purge(), Job.remove()
```
koresar, line 9: Is this comment correct? I don't see that in the PR changes.

koresar (changes requested):
Code is good. Some code comments needs rewording though.

pdfowler:
@koresar Thanks for the feedback and callouts on the comments/docs. I've updated my code with the upstream master and migrated my code to Typescript.

Only remaining task is the HTML docs, which I believe are handled as a part of the release process via the `docs` script. Let me know if there's anything else I can do on this PR

koresar (approved):

koresar:
Quality PR! Thanks for that.

koresar:
Published in v4.2.0.
Thanks for your contribution!

Issue #1118, opened 2020-09-15
Agenda doesn't seem to be honoring lock limits


GimpMaster:
I have 4 of the same Agenda node processes running on a single machine using pm2 for process management.

Each of the 4 job runners is expected to take tasks on the start of each hour and compute data for a long duration (multiple minutes). To make sure that no single process locks all the jobs on the start of the hour I set on the Agenda initilization a number of lock limits as seen below:

``` javascript
let agenda = new Agenda({
	db: {
		address: connectionString,
		options: { useUnifiedTopology: true },
		collection: 'agenda'
	},
	maxConcurrency: 1,
	defaultLockLimit: 1,
	defaultConcurrency: 1,
	lockLimit: 1
});
```

Watching the output logs from all 4 processes it seems as though one Agenda is locking all jobs and then working through them sequentially instead of each Agenda grabbing one job, processing, and then returning for the next job.

Any suggestions?

GimpMaster:
Using v3.1.0

GimpMaster:
Looking through the agenda code I believe the problem lives here:

``` javascript
  async function jobQueueFilling(name) {
    // Don't lock because of a limit we have set (lockLimit, etc)
    if (!shouldLock(name)) {
      debug('lock limit reached in queue filling for [%s]', name);
      return;
    }

    // Set the date of the next time we are going to run _processEvery function
    const now = new Date();
    self._nextScanAt = new Date(now.valueOf() + self._processEvery);

    // For this job name, find the next job to run and lock it!
    try {
      const job = await self._findAndLockNextJob(name, definitions[name]);
      // Still have the job?
      // 1. Add it to lock list
      // 2. Add count of locked jobs
      // 3. Queue the job to actually be run now that it is locked
      // 4. Recursively run this same method we are in to check for more available jobs of same type!
      if (job) {
        debug('[%s:%s] job locked while filling queue', name, job.attrs._id);
        self._lockedJobs.push(job);
        definitions[job.attrs.name].locked++;
        enqueueJobs(job);
        await jobQueueFilling(name);
        jobProcessing();
      }
    } catch (error) {
      debug('[%s] job lock failed while filling queue', name, error);
    }
  }

```

Specifically what I think is happening is Agenda calls processJobs on a setInterval(). For every single job definition it calls jobQueueFilling(name), let's say there are 10 definitions. 

[...]

PR #1119, opened 2020-09-16
Agenda being too greedy on locking jobs

GimpMaster:
- Fixed issue with too many locks being set asynchronously

### DIFF
--- a/lib/utils/process-jobs.js
+++ b/lib/utils/process-jobs.js
@@ -170,6 +170,14 @@ module.exports = function(extraJob) {
       // 3. Queue the job to actually be run now that it is locked
       // 4. Recursively run this same method we are in to check for more available jobs of same type!
       if (job) {
+        // Before en-queing job make sure we haven't exceed our lock limits
+        if (!shouldLock(name)) {
+          debug('lock limit reached before job was returned. Releasing lock on [%s]', name);
+          job.attrs.lockedAt = null;
+          self.saveJob(job);
+          return;
+        }
+
         debug('[%s:%s] job locked while filling queue', name, job.attrs._id);
         self._lockedJobs.push(job);
         definitions[job.attrs.name].locked++;


GimpMaster:
Fixes issue #1118

simllll (approved):
there is potentially one issue with this approach,
when calling saveJob it is checked if the nuxtRunAt is lower then the _nextScanAt (basically now). If it is so, it calls immediately job to run again. It is not a deal breaker, but it could lead to more cpu power wasted than needed. But this is also a possible problem on other situations, therefore good to go for me!

in the future we can think of a generic "unlock" method, that just unlocks the job in the db and discards the job locally. (see for example unlockJob in DBRepository https://github.com/hokify/agenda/commit/a70f5009edd4d689305da6381caa08fec9c37036)

GimpMaster:
Thank you @simllll . I noticed that this branch has merge conflicts. Are you wanting me to fix that or is that something you guys fix during the merge process.

simllll:
Right now there is a transformation to typescript going on, but there is not any specific roadmap or time plan. Therefore I think the best way to cope with this situation is, if I ping you when the master is ready for bug fixes again? If this is okay for you? Thanks a lot!

GimpMaster:
Yep that works for me. Thank you

simison:
@GimpMaster hi! Thanks for patience. I think we're little more stable again so this could be fine for rebase. :-)

GimpMaster:
@simison - Sure thing. Give me a day or so to get back to it.

leonardlin:
@GimpMaster @simllll 
Sorry to jump on this late.

In the issue ticket: https://github.com/agenda/agenda/issues/1118

you mention:


> Each time jobQueueFiling is called it tests to see if it should lock that job, on the first go you have Zero lock jobs and thus it passes the shouldLock() test. Then it fires off an await self.FindAndLockNextJob(). While Javascript is waiting for the "await" to return the process loop continues and it does this same await self.FindAndLockNextJob() call which will effectively lock ALL 10 jobs immediately by a single process.

Do I understand you correctly that you think:

while await self.FindAndLockNextJob() is waiting, the loop continues and jobQueueFilling is executed 10 times, until await self.FindAndLockNextJob() returns no job?

`await` should not continue at this point but simply wait.
What I can however imagine is that there are multiple `jobQueueFilling` running in parallel or `lockOnTheFly` might mix in here too.

PR https://github.com/agenda/agenda/pull/1086 is dealing also with "lockLimits" not honored.

if you guys ok, I'd like to reproduce this issue with and without https://github.com/agenda/agenda/pull/1086 fix in.
Perhaps it's the same issue.

simllll:
Yes, I have seen the other Pr today already.. looks for me like the same bug we are talking here!

leonardlin:
@GimpMaster 
With multiple agenda-processor, do you mean something like this?

`
const agenda1 = new Agenda();
const agenda2 = new Agenda();
const agenda3 = new Agenda();
const agenda4 = new Agenda();
agenda1.start();
agenda2.start();
agenda3.start();
agenda4.start();
`

or you start 4 separate node - processes albeit via pm2-library
`
$ node app.js
$ node app.js
$ node app.js
$ node app.js
`

GimpMaster:
@leonardlin - The second one you presented. 4 separate node app.js being run by a pm2 manager.

Since switching to my fork I have had zero issues for the past 2 months running 24 hours a day with 4 unique processes. No locking issues.

I will try to fix merge conflicts tomorrow night. Work has been busy.

leonardlin:
@GimpMaster  let me know if you need a hand, no worries if you are too busy with work. I can try to help you with it. Just let me know.

GimpMaster:
Sorry guys - @leonardlin . I think this is ready now. I pulled agenda master into my fork and seems like there are no issues. Let me know.

PR #1120, opened 2020-09-19
fix(save-job): add job name to db queries to allow mongodb sharding

simllll:
this allows using sharded collections for improved performance, as this is the only place where job name is not provided.

sharding by name:
db.runCommand( { shardCollection: "YOUR-COLLECTION.agendaJobs", key: { name: 1 } } )


koresar (dismissed):
LGTM

simllll:
this pr was opened from a master branch, which has changed significantly in the meantime.

PR #1123, opened 2020-10-07
WIP: Rewrite in Typescript

simllll:
Hi all,
this is the initial version of agenda written in typescript. I'm asking for comments, reviews and how we should handle the future of it! 

I tried to refactor everything and add types. This is just an initial version, it should run, but not all tests complete yet.

To clarify what's in here:
- a complete rewrite (I tried keeping git history, but this didn't really work out). Too many files and code logic has changed.
- introduced a DB Adapter Layer, where all the database calls are happening (JobDbRepository) 
- typed everything I could type for now to have type safety in TS and can also export it
- Worked on the test, converted them to ts too and they are using MongoMemoryServer now (this fails on travis right now though, works locally)
- Added eslint&prettier
- Breaking changes (e.g. define(name, processor, options?): options moved from 2nd to 3rd parameter,... )
- touch() can have a progress parameter which is saved to the job

Why?
I think agenda is a great tool, I tried bull and others, but nothing came to the services that agenda offers. But current agenda code base is hard to understand and therefore also hard to maintain.

What now?
Help! Let us fix all tests and improve agenda even more. This is just a inital refactoring, I stumpled on a lot of things that could be improved in the future.

Future?
Exactly, unsure about this.. should we merge it into the agenda repo or should I create a fork (e.g. agenda.ts).. What do you think guys? @niftylettuce @koresar @simison @MichielDeMey @cjolif @rschmukler @OmgImAlexis  & everyone I missed here :)

Todo:
- [x] fix tests
- [ ] describe changes
- [ ] update docs

PR #1129, opened 2020-10-14
Feature/switch to ts

OmgImAlexis:
This is a more focused approach to what https://github.com/agenda/agenda/pull/1123 was trying to achieve.

All tests are passing.

Yes this does contain quite a few `any` types and a few `// @ts-expect-error` lines but those can be fixed later on. For now let's just get this merged.


OmgImAlexis:
I haven't updated the lock file as I don't use yarn myself. Maybe for the benefit of new users we should switch to a `package-lock` so they don't need to install yarn?

test/job.js
```diff
@@ -864,20 +864,21 @@ describe('Job', () => {
       await agenda.stop();
     });
 
-    it('does not on-the-fly lock more than definition.lockLimit jobs', async() => {
-      agenda.define('lock job', {lockLimit: 1}, (job, cb) => {}); // eslint-disable-line no-unused-vars
+    // @todo fix this test
```
OmgImAlexis, line 867: This test currently fails. I'd suggest we get it fixed before merging this in.

OmgImAlexis:
I've disabled linting when running the `test` script for now as I'll need to reconfigure xo correctly for ts.

lib/job/schedule.ts
```diff
@@ -1,17 +1,15 @@
-'use strict';
-const date = require('date.js');
+// @ts-expect-error
```
simison, line 1: Nit; usually helpful to add extra comment to elaborate why error/warning was suppressed, even if it seems obvious.

lib/agenda/cancel.ts
```diff
@@ -1,18 +1,19 @@
-'use strict';
-const debug = require('debug')('agenda:cancel');
+import { Agenda } from './index';
```
simllll, line 1: as you only use the type from Agenda here, you can use
"import type { Agenda } from './index'"
this helps for detecting circular dependencies.. index is importing cancel too. By just importing the type, circular dependency checks are not a problem.

lib/utils/process-jobs.ts
```diff
@@ -63,16 +63,14 @@ module.exports = function(extraJob) {
 
   /**
    * Internal method that adds jobs to be processed to the local queue
-   * @param {*} jobs Jobs to queue
-   * @param {boolean} inFront puts the job in front of queue if true
-   * @returns {undefined}
+   * @param jobs Jobs to queue
    */
-  function enqueueJobs(jobs) {
+  function enqueueJobs(jobs: any[] | any) {
```
simison, line 68: Why not type `Job` here instead of `any`?

tsconfig.json
```diff
@@ -0,0 +1,72 @@
+{
+  "include": [
+    "lib"
+  ],
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",                          /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
```
simllll, line 10: agenda targets node 10 as minimum requirement if I recall correctly,
therefore
```
		"module": "commonjs",
		"target": "es2018",
		"lib": ["es2018"],
```
should be fine

i always woudl recommend adding
		"declaration": true

lib/utils/process-jobs.ts
```diff
@@ -206,6 +204,7 @@ module.exports = function(extraJob) {
       debug('[%s:%s] nextRunAt is in the past, run the job immediately', job.attrs.name, job.attrs._id);
       runOrRetry();
     } else {
+      // @ts-expect-error
```
simison, line 207: Nit; I'd elaborate why this was suppressed here, and in other places where used.

test/agenda.js
```diff
@@ -3,9 +3,9 @@
 const expect = require('expect.js');
 const {MongoClient} = require('mongodb');
 const delay = require('delay');
-const Job = require('../lib/job');
-const hasMongoProtocol = require('../lib/agenda/has-mongo-protocol');
-const Agenda = require('..');
+const { Job } = require('../dist/job');
```
simllll, line 6: tests should include src files if possbile.
to enable mocka to run typescript files, just add -r ts-node/register  as command line parameter:
```
"mocha": "mocha -r ts-node/register --reporter spec --timeout 8000 -b",
    "mocha-debug": "DEBUG=agenda:**,-agenda:internal:** mocha -r ts-node/register --reporter spec --timeout 8000 -b",
    "mocha-debug-internal": "DEBUG=agenda:internal:** mocha -r ts-node/register --reporter spec --timeout 8000 -b",
    "mocha-debug-all": "DEBUG=agenda:** mocha -r ts-node/register --reporter spec --timeout 8000 -b",
```
and also check Makefile, see
https://github.com/hokify/agenda/blob/typescript/Makefile
https://github.com/hokify/agenda/blob/typescript/package.json
for reference

package.json
```diff
@@ -10,7 +10,9 @@
     "node": ">=10.0.0"
   },
   "scripts": {
-    "test": "npm run lint && npm run mocha",
+    "build": "tsc",
+    "pretest": "npm run build",
+    "test": "npm run mocha",
```
simison, line 15: Why not run `lint` here too?

lib/utils/process-jobs.ts
simllll, line 68: it should be a job that is added here.
```
import type { Job } from '../job';
jobs: Job[] | Job

```

lib/utils/process-jobs.ts
```diff
@@ -63,16 +63,14 @@ module.exports = function(extraJob) {
...
+   * @param jobs Jobs to queue
    */
-  function enqueueJobs(jobs) {
+  function enqueueJobs(jobs: any[] | any) {
     if (!Array.isArray(jobs)) {
       jobs = [jobs];
     }
 
-    jobs.forEach(job => {
+    jobs.forEach((job: any) => {
```
simllll, line 73: Job

docker-compose.yml
```diff
@@ -1,9 +1,9 @@
-# Run tests in Docker `docker-compose up` on Node.js v4
+# Run tests in Docker `docker-compose up` on Node.js v14
```
simison, line 1: Might be safe to just drop the version in comment since it's hard to keep updated. :-)

lib/agenda/define.ts
```diff
@@ -1,17 +1,18 @@
...
- * @param {String} name name of job
- * @param {Object} options options for job to run
- * @param {Function} processor function to be called to run actual job
- * @returns {undefined}
+ * @param name name of job
+ * @param options options for job to run
+ * @param processor function to be called to run actual job
  */
-module.exports = function(name, options, processor) {
+export const define = function(this: Agenda, name: string, options: any, processor: Function) {
```
simison, line 15: Can we have types for `options`?

simllll (commented):
I'm currently quite busy, just some first comments, if you'd like to I will review in more detail later. Let me know if you like me to add more types?

lib/agenda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+  _mongoUseUnifiedTopology?: boolean;
+  _name: any;
+  _processEvery: number;
+  _ready: Promise<unknown>;
+  _sort: any;
+  _db!: MongoClient;
+  _mdb!: MongoDb;
+  _collection: any;
+  _nextScanAt: any;
+  _processInterval: any;
```
simison, line 73: Can some of these `any`'s be `number`?

lib/agenda/index.ts
```diff
@@ -0,0 +1,156 @@
...
+ * @property {Boolean} _isLockingOnTheFly
+ * @property {Array} _jobsToLock
+ */
+class Agenda extends EventEmitter {
+  _defaultConcurrency: any;
+  _defaultLockLifetime: any;
+  _defaultLockLimit: any;
+  _definitions: any;
+  _findAndLockNextJob: any;
+  _indices: any;
```
simison, line 56: Would it be possible to define object shape here?

```js
this._indices = {name: 1, ...this._sort, priority: -1, lockedAt: 1, nextRunAt: 1, disabled: 1};
```

lib/job/schedule.ts
OmgImAlexis: These aren’t meant to be kept. They should all be removed in the next PR we open for this.

OmgImAlexis:
I might as well just close this. No one is reading.

This PR is the start is the transfer to ts. If we end up spending a load of time working on fixing all the little things this PR will get too big and won’t be merged. Let’s merge this and then get those issues fixed.

test/agenda.js
OmgImAlexis: No. Ts node isn’t the same. We should be testing what users get.

lib/agenda/cancel.ts
OmgImAlexis: Circular dependencies aren’t an issue when typescript compiles. 🙄

simllll: I don't understand your down vote.
Yeah it is not a general problem, it is just a more defined approach how you import and use "Agenda".
Please don't get offended, I just was offering my help!

test/agenda.js
simllll: Wouldn't fully agree with this.
There are two approaches, test early, test late.
Right now (in the old code), as the source file is the same as the packaged file, it is a test early approach. Which means: testing the source files. There are other tests that guarantee that source files are transpiled correctly. Or other tests to see if package building etc works correctly. Therefore sticking to test early would make sense to me. But as always, there are definitely more opinions about this ;) So just my two cents :+1:

simison:
>f we end up spending a load of time working on fixing all the little things this PR will get too big and won’t be merged. Let’s merge this and then get those issues fixed.

@OmgImAlexis sorry, I definitely glossed over the PR description! I reviewed the first thing I woke up from phone, so prone to do so. :-) 

I should've clarified that none of my comments are meant blocking, just questions that are totes fine addressed later!

Would both you and @simllll be comfortable merging this as-is since tests are passing as-is, and working on follow-ups?

I'm not very proficient in TypeScript so I won't be able to help with much more than questions and nitpicks anyway. ;-)

So, I think you're right, I'd be happy to see this merged as-is.

test/agenda.js
OmgImAlexis: This is nothing todo with that. Typescript, ts node and compiled JavaScript are all different in their own ways. Adding typescript issues into the mix isn’t going I help devs.

If we need source maps they can be enabled for testing.

simison (commented):
Thanks for starting on this!

I think it would be fine to mark the failing test as skipped and follow-up with separate PR.

simllll:
> Would both you and @simllll be comfortable merging this as-is since tests are passing as-is, and working on follow-ups?

What do you think about creating a seperate branch where we open all PRs against it? e.g. "v4" or "typescript". So we can seperate all the parts in own PRs, but never will have a code that is not fully working or not finsihed in master?

But just an idea, I'm also okay with merging this into the master and keep on working on it from there on :)

lib/agenda/index.ts
Uzlopak: Actually no, I think collection is a mongodb connection, nextScanAt is null or a Date, and processInterval is string or a number as processInterval can be afaik a number, a chron string or human readable string like "5 seconds"

lib/agenda/cancel.ts
Uzlopak: import type is the way to go. When transpiled to javascript the import will be dropped as it is marked as type only import.

lib/agenda/define.ts
Uzlopak: Not really as you would need more interfaces and this PR tries the minimalistic approach

lib/agenda/index.ts
simllll: _processInterval is actually a timer .. therefore `_processInterval?: ReturnType<typeof setInterval>` would be correct.  _processEvery is the thing you were referring too, it's a string or a number.

feel free to look up types in my fork, I spended already some time to figure them out :-)
https://github.com/hokify/agenda/tree/typescript/src/types
https://github.com/hokify/agenda/blob/typescript/src/index.ts

lib/agenda/cancel.ts
OmgImAlexis: If setup correctly typescript it can detect that without the need for the explicit “type import”. We shouldn’t be replying on a dev to add “type” before an import to solve this instead we should setup our config correctly.

tsconfig.json
OmgImAlexis: I don’t believe this is needed once we switch to tslib.

lib/agenda/index.ts
OmgImAlexis: Wrapping a type in a generic like ReturnType makes it harder to read the actual type in an IDE I’d suggest against them.

OmgImAlexis:
@simison is there a reason we don’t have rebase merging enabled? Was going to merge this but all that’s allowed is squash.

lib/agenda/index.ts
simllll: It's an internal type, everything is better than any ;)

simison:
> is there a reason we don’t have rebase merging enabled? Was going to merge this but all that’s allowed is squash.

Just to guide folks to do the right thing which most of the time is squashing; backlog is especially important for packages like this which lots of people use and check for changes. If you need a rebase-merge, just toggle the setting temporarily on. :-)

lib/agenda/cancel.ts
simllll: That's not true, no idea where you found this information. It has nothing todo with the setup itself, if you include a class file, there are just many cases where a compiler can not figure out if it is a type only import or not. By specifying import type you not just give the compiler the right hint,but also forbid non type usage in the file. It's super useful, look it up: https://github.com/microsoft/TypeScript/pull/35200

OmgImAlexis: Yes using tsc itself does that. We’re not going to be using tsc.

simllll: Can you shed some light on this statement? How you gonna use typescript without typescript compiler? ;) And besides that, no offence, but it would be great to let things open to discuss, instead of deciding on the best way ahead. At least that's what is my experience with open source projects and collaboration ;)

lib/job/schedule.ts
Uzlopak: it is even the wrong comment operator. @ts-expect-error was designed to be used for unit tests, where you actually have type errors because you e.g. give incomplete objects violating the typings. This has to be actually an @ts-ignore

OmgImAlexis: No it doesn’t. Expect error can be used when you have code that’s correctly throwing a type issue and you expect it to stay that way for now. If we change any other code which solves this then this will error and the dev will know to remove it.

lib/agenda/cancel.ts
OmgImAlexis: `tsc` is **A** compiler.

We'll be using https://github.com/egoist/tsup which will give us a es5 commonjs bundle with a definition file. The commonjs won't include the import as it's not used for anything but the type.

lib/agenda/index.ts
OmgImAlexis: Using internal generics doesn't make it easier. It means that new dev not only now need to know typescript but also generics, etc. Just use the type. It's a number that `setInterval` returns, yes this varies in browsers, etc. we're writing a node lib so that doesn't matter.

OmgImAlexis: Screen Shot 2020-10-15 at 7 46 05 pm


For inexperienced devs this is going to be way more confusing seeing it show a `NodeJS.Timeout` on a setInterval especially when `ReturnType<typeof setTimeout>` also returns that type. See why generics aren't always helpful, yes technically it's returning the "correct" type but that doesn't mean it's the best one for devs to see.

lib/agenda/cancel.ts
simllll: typescript and esbuild is not recommended for produciton (yet). They haven't reached 1.x. https://github.com/evanw/esbuild#disclaimers

But open to new tools, especially because esbuild is super fast :) But what would be a valid reason to use a so fresh builder which hasn't reached a stable release yet?

Besides that, it wouldn't solve the issue with import type ;-)

OmgImAlexis: Welcome to open source, this isn't enterprise we can use tools that aren't perfect.

> Besides that, it wouldn't solve the issue with import type ;-)
And yet it does... I literally use this at work. Please stop trying to state facts on software you don't use.

PR #1131, opened 2020-10-14
fix(tests): wrong initial value and lower processEvery for agenda-instance.js

simllll:
- the initial value is probably just a type mistake that is there since the initial commit. 
- the second one is actually a side effect of another bug that it is working currently, but processEvery has a default value of 5 seconds, therefore only every 5 seconds it is checked if there are new jobs available. The tests itself wait just for 400ms before sending back that the job didn't run..which was not even possible, because agenda could easily be still waiting for 5 seconds to be passed.


simison (approved):
Good catch! Thanks.

PR #1138, opened 2020-10-20
fix: unlock job test needs cb

simllll:

PR #1142, opened 2020-10-31
Revert "fix: unlock job test needs cb"

simison:
Reverts agenda/agenda#1138

This worked somehow on PR but now I'm seeing consistent failures on other rebased PRs for this test:

> does not on-the-fly lock more than agenda._lockLimit jobs

FYI @ simllll

### DIFF
--- a/test/job.js
+++ b/test/job.js
@@ -642,8 +642,8 @@ describe('Job', () => {
     });
 
     it('clears locks on stop', async() => {
-      agenda.define('longRunningJob', (job, cb) => { // eslint-disable-line no-unused-vars
-        // Job never finishes (the 2nd parameter is important, otherwise the job finishes immediately)
+      agenda.define('longRunningJob', job => { // eslint-disable-line no-unused-vars
+        // Job never finishes
       });
       agenda.every('10 seconds', 'longRunningJob');
       agenda.processEvery('1 second');


simllll:
Mhm, are the test seems right though? Are you sure this is the real issue here? I've fixed the test in my fork the same way and test suite is super stable?

simison:
@simllll it for sure is odd! I didn't have time yet to dig deeper but I rebased two Renovate PRs and I can see same test fail on both now:

https://github.com/agenda/agenda/pull/1141
https://github.com/agenda/agenda/pull/1134

As well in this current PR.

Any thoughts on what it might be?

simllll:
No idea what exactly happened here, but the test has been changed by @OmgImAlexis , I remember something while changing the code to ts files.

I've added already several new tests in my fork, I will try to find some time to bring them in here too, because there are actually some bugs in the current code base which we should tackle ;) Especially with job locking, concurrency and lock limits. I guess this "randomly" occuring bug has something todo with this too.

Tried to debug it quickly, the test is right, lockLimit is not honored. There are two jobs locked, but there should only be one. Couldn't locate the real issue yet.

simison:
> I will try to find some time to bring them in here too

That would be great! 🎉

simison:
@simllll I'll close this one since the issue seemed to be elsewhere. :-)

PR #1150, opened 2020-12-04
Migrate cron to cron-parser

leonardlin:
Changes:
* added type declaration for return in `stop.ts`
* migrated cron to cron-parser based on old PR: https://github.com/agenda/agenda/pull/475
* exposed mangodb, that it's easier to access from external
* added 2 tests for cron patterns
** Testing from 1-based cron-month
** Testing for reading of cron pattern

Breaking Change:
* cron package used a non-standard cron pattern. months started at 0. Standard for unix would be to start at 1. This is also mentioned as an issue on the cron issue page: https://github.com/kelektiv/node-cron/issues/475

Comments:
* Thanks to original PR, https://github.com/agenda/agenda/pull/475
* I hope this will solve cron-issues, let's clean up all PRs/Issues related to cron and retest afterwards


lib/job/compute-next-run-at.ts
```diff
@@ -36,14 +36,16 @@ export const computeNextRunAt = function(this: Job) {
     debug('[%s:%s] computing next run via interval [%s]', this.attrs.name, this.attrs._id, interval);
     let lastRun = this.attrs.lastRunAt || new Date();
     lastRun = dateForTimezone(lastRun);
+    var cronOptions: any = { currentDate: lastRun.toDate() };
```
simison, line 39: Nit, could you switch this `var` to `const` to keep code consistent? Thanks!

leonardlin: Good catch, I will fix the PR.

simison:
Thanks for adding the tests! Much appreciated.

Not a blocker for merging, but it would help us if you can write quick "breaking" blurb under "Next" section in `History.md` and instructions how to upgrade from old behavior.

leonardlin:
> 
> 
> Thanks for adding the tests! Much appreciated.
> 
> Not a blocker for merging, but it would help us if you can write quick "breaking" blurb under "Next" section in `History.md` and instructions how to upgrade from old behavior.

sure thing, will add

simison (approved):
Thank you!

test/agenda.js
```diff
@@ -236,6 +236,12 @@ describe('Agenda', function() { // eslint-disable-line prefer-arrow-callback
         it('returns a job', async() => {
           expect(await jobs.every('5 minutes', 'send email')).to.be.a(Job);
         });
+        it('cron job with month starting at 1', async() => {
+          expect(await jobs.every('0 0 * 1 *', 'send email').then(({attrs}) => attrs.nextRunAt.getMonth())).to.be(0); // javascript getMonth() return 0 for january
+        });
+        it('repeating job with cron', async() => {
+          expect(await jobs.every('*/5 * * * *', 'send email').then(({attrs}) => attrs.nextRunAt)).to.not.be(null);
+        });
```
simison, line 244: @leonardlin just thought of this — definitely no need to jump to this but what do you think about adding a test for timezone config?

PR #1169, opened 2020-12-19
lint fixes 1 - tests

leonardlin:
lint fixes part 1, tests

### DIFF
--- a/test/job.js
+++ b/test/job.js
@@ -766,14 +766,13 @@ describe('Job', () => {
       let startCounter = 0;
 
       const processorPromise = new Promise(resolve => {
-        agenda.define('lock job', { lockLifetime: 50 }, () => {
+        agenda.define('lock job', { lockLifetime: 50 }, async() => {
           startCounter++;
 
           if (startCounter !== 1) {
             expect(startCounter).to.be(2);
-            return agenda.stop().then(() => { // eslint-disable-line promise/prefer-await-to-then
-              resolve();
-            });
+            await agenda.stop();
+            resolve();
           }
         });
       });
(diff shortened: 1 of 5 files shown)


test/job.js
```diff
@@ -765,21 +765,18 @@ describe('Job', () => {
...
-        }, async() => {
+      const processorPromise = new Promise(resolve => {
+        agenda.define('lock job', { lockLifetime: 50 }, () => {
           startCounter++;
 
           if (startCounter !== 1) {
             expect(startCounter).to.be(2);
-            await agenda.stop();
-            resolve();
+            return agenda.stop().then(() => { // eslint-disable-line promise/prefer-await-to-then
```
simison, line 774: How come this needed `.then()`? 🤔 

Would be good to add a code comment to explain lint disable.

You might also shorten it to `.then(resolve)`

simison (dismissed):
Looking good!

Overall I prefer to have a brief comment on why a lint disable was needed for other devs reading the code later on, but not a blocker for merging.

test/job.js
```diff
@@ -794,21 +790,18 @@ describe('Job', () => {
...
-        }, async(job, cb) => { // eslint-disable-line no-unused-vars
+      const processorPromise = new Promise(resolve => {
+        agenda.define('lock job', { lockLifetime: 50 }, (job, cb) => { // eslint-disable-line no-unused-vars
           runCount++;
 
           if (runCount !== 1) {
             expect(runCount).to.be(2);
-            await agenda.stop();
-            resolve();
+            agenda.stop().then(() => { // eslint-disable-line promise/prefer-await-to-then
```
simison, line 799: Another likely not needed `.then()` here.

test/job.js
leonardlin: changed to await style

test/job.js
leonardlin: changed to await style

simison (approved):
Thanks!

PR #1170, opened 2020-12-20
lint fixes 2 - utils

leonardlin:
lint fixes in the utils folder.
There were dangling promises. Please take a look at them.
All tests pass but to be certain.


simison (approved):
Thanks for isolating this to a more focused, small PR. 👍 

I haven't tested but it's looking good.

PR #1171, opened 2020-12-20
lint fixes 3 - lib/agenda

leonardlin:
mainly lint fixes in lib/agenda.

There will be 4th package with lint fixes

### DIFF
--- a/lib/agenda/create.ts
+++ b/lib/agenda/create.ts
@@ -1,6 +1,6 @@
 import createDebugger from 'debug';
 import { Job } from '../job';
-import { Agenda } from './index';
+import { Agenda } from '.';
 
 const debug = createDebugger('agenda:create');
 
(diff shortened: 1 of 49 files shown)


lib/agenda/create.ts
```diff
@@ -11,9 +11,9 @@ const debug = createDebugger('agenda:create');
  * @param name name of job
  * @param data data to set for job
  */
-export const create = function(this: Agenda, name: string, data: {}): Job {
+export const create = function(this: Agenda, name: string, data: any): Job {
```
simison, line 14: How come `any` and not `object`? (and maybe it can also be `null`?)

lib/agenda/jobs.ts
```diff
@@ -11,9 +12,9 @@ import { createJob } from '../utils';
  * @param number of documents to skip in MongoDB
  * @returns resolves when fails or passes
  */
-export const jobs = async function(this: Agenda, query = {}, sort = {}, limit = 0, skip = 0) {
+export const jobs = async function(this: Agenda, query: FilterQuery<any> = {}, sort = {}, limit = 0, skip = 0) {
   const result = await this._collection
-    .find(query)
+    .find(query) // eslint-disable-line unicorn/no-fn-reference-in-iterator
```
simison, line 17: Just for reference for anyone reading, since I was curious myself what was happening here: https://github.com/sindresorhus/eslint-plugin-unicorn/issues/755

simison (approved):
👍 

Needs a rebase, but looking good. I have not tested this locally.

lib/agenda/create.ts
leonardlin: yeah, I was surprised too.
The unicorn-lint-styles do not allow 'object' but allows 'any'.

In newer Typescript there is also a Record<string, type> - type, but I'm not too familiar with it, so I didn't try that one.

simison: Aight! Not a biggie anyway, I don't think it needs to block the PR.

simison (approved):
Thanks for the cleanup! 👍

PR #1178, opened 2020-12-28
[types] fix agenda.<mongo> types bug

### LINKED ISSUE #1156: TypeScript error Argument of type 'Db' is not assignable to parameter of type 'MongoClient' because Agenda.mongo parameter is declared as MongoClient type instead of Db type
Correct me if i'm wrong, but it is mentioned in official documentation that we can `Use an existing mongodb-native MongoClient/Db instance for the mongo`  https://github.com/agenda/agenda/issues/868 

We notice that many Db functions are used with `_mdb`, and often it is treated as a Db instance.
This means that the `mdb` parameter of the Agenda.mongo function should be of the Db type.

See this line:

https://github.com/agenda/agenda/blob/fba94e7fc32b1ceb0ba68643adf04055109ddefc/lib/agenda/mongo.ts#L8

Also, there is a `ts-expect-error` on line 13, which means that an error is being ignored in TypeScript.

https://github.com/agenda/agenda/blob/fba94e7fc32b1ceb0ba68643adf04055109ddefc/lib/agenda/mongo.ts#L13

Because of this declaring of the `mdb` parameter as the `MongoClient` type instead of `Db` type, the types definition for Agenda recently got updated in the file linked below and now we face TypeScript errors whenever we pass in a Db with the `Agenda.mongo` function.

https://github.com/max003003003/DefinitelyTyped/blob/dd4d8083a606daed6a27cb77aad4c74737f4e42b/types/agenda/index.d.ts#L54

In our own project, we currently face this error when we pass a Db argument to `Agenda.mongo`, which has been working all along and it is meant to be used this way even according to official documentation:

Our code: 

```js
await agenda.mongo(dbObject, "Jobs").start();
```

Error:

```
Argument of type 'Db' is not assignable to parameter of type 'MongoClient'.
  Type 'Db' is missing the following properties from type 'MongoClient': connect, close, db, isConnected, and 3 more
```

erfanium:
fixes #1156

### DIFF
--- a/lib/agenda/mongo.ts
+++ b/lib/agenda/mongo.ts
@@ -1,4 +1,4 @@
-import { Collection, MongoClient } from 'mongodb';
+import { Collection, Db } from 'mongodb';
 import { Agenda } from '.';
 
 /**
@@ -9,8 +9,7 @@ import { Agenda } from '.';
  * @param collection name collection we want to use ('agendaJobs')
  * @param cb called when MongoDB connection fails or passes
  */
-export const mongo = function(this: Agenda, mdb: MongoClient, collection: string, cb?: (error: Error, collection: Collection<any> | null) => void): Agenda {
-  // @ts-expect-error
+export const mongo = function(this: Agenda, mdb: Db, collection: string, cb?: (error: Error, collection: Collection<any> | null) => void): Agenda {
   this._mdb = mdb;
   this.db_init(collection, cb);
   return this;


leonardlin (approved):

leonardlin:
sorry, my bad. I thought the change is trivial, but I saw that it's not the type that was wrong but the assignment

erfanium:
@leonardlin
I didn't clearly understand what you meant, but ok😁

leonardlin:
@erfanium  sorry please ignore. I thought I merged THE PR by mistake. But it's all good.

PR #1179, opened 2020-12-28
Fix "[types] fix agenda.<mongo> types bug"

leonardlin:
Reverts agenda/agenda#1178 and makes correction


leonardlin:
opened by mistake

PR #1182, opened 2020-12-28
CI: add lint check

simison:
- Adds a linting task to Travis CI. Runs it only on one of the containers to avoid running it 3 times.
- Fix few lint issues

I didn't look too deep if there's a simpler way to run this, here are docs if anyone wants to simplify this further:

https://docs.travis-ci.com/user/build-matrix/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,18 @@
 language: node_js
+node_js:
+  - 10
+  - 12
+  - 14
 services: mongodb
 sudo: required
-matrix:
-  fast_finish: true
+jobs:
   include:
-    - name: Node.js 10
-      node_js: 10
-      env: RUN="yarn test"
-    - name: Node.js 12
-      node_js: 12
-      env: RUN="yarn test"
-    - name: Node.js 14 & lint
-      node_js: 14
-      env: RUN="yarn lint && yarn test"
+    - stage: lint
+      name: "Check Linting"
+      script: "yarn lint"
+    - stage: test
+      name: "Test"
+      script: "yarn test"
 addons:
   apt:
     sources:
@@ -31,7 +31,6 @@ env:
     # NodeJS v4+ requires gcc 4.8+
     - CXX=g++-4.9 CC=gcc-4.9
     - TRAVIS=true
-script: '$RUN'
 before_script:
   - mongo --version
   - mongod --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ matrix:
       env: RUN="yarn test"
     - name: Node.js 14 & lint
       node_js: 14
-      env: RUN="yarn test && yarn lint"
+      env: RUN="yarn lint && yarn test"
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,18 @@
 language: node_js
 services: mongodb
 sudo: required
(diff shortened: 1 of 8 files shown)


lib/agenda/index.ts
```diff
@@ -101,7 +101,7 @@ class Agenda extends EventEmitter {
     super();
 
     this._name = config.name;
-    this._processEvery = (humanInterval(config.processEvery) ?? humanInterval('5 seconds')) as number;
+    this._processEvery = (humanInterval(config.processEvery) ?? humanInterval('5 seconds')) as number; // eslint-disable-line @typescript-eslint/non-nullable-type-assertion-style
```
simison, line 104: This was noted also at https://github.com/agenda/agenda/pull/361#pullrequestreview-559199264

Wasn't quite sure how to go about this either.

.travis.yml
```diff
@@ -1,10 +1,18 @@
...
+  include:
+    - name: Node.js 10
+      node_js: 10
+      env: RUN="yarn test"
+    - name: Node.js 12
+      node_js: 12
+      env: RUN="yarn test"
+    - name: Node.js 14 & lint
+      node_js: 14
+      env: RUN="yarn test && yarn lint"
```
leonardlin, line 15: really minor thought.

'linting' is often checked before 'test'

That's something developers forget often and it fails really fast.

leonardlin (commented):
looks good to me.
There is a merge conflict, but otherwise I think good to go.

.travis.yml
simison: Fair point! Changed in https://github.com/agenda/agenda/pull/1182/commits/15872e77c4b17f42a7c6f4e7b552fc0f317127fc

leonardlin:
this is weird.
yarn lint seems to take '&&' as parameter, it causes the build to pass, although it should have failed.

image

.travis.yml
```diff
@@ -1,10 +1,18 @@
 language: node_js
 services: mongodb
 sudo: required
-node_js:
-  - 10
-  - 12
-  - 14
+matrix:
```
koresar, line 4: I've seen few times that linting can actually fail under one node.js version and pass on another.
I'd recommend removing the "avoid running it 3 times". Let's lint on each node.js.

Also, I'd recommend us to do more KISS principle. Simplify not for machines but for humans (us). In other words, this was a premature optimisation. Let's remove the `matrix` entirely.

Also, as it was reported on a screenshot, the tests do not actually run under node 14.

koresar (changes requested):
I recommend a simplification of the GitHub Actions configuration.

.travis.yml
simison: Yep, fair points! Feel free to modify unless I get to it sooner. :-)

leonardlin:
Simplified build.
Introduced a lint-error to test if travis woudl fail: https://travis-ci.org/github/agenda/agenda/builds/754085260
this worked.

leonardlin (approved):

.travis.yml
leonardlin: @koresar take a look, I updated the travis config.
I tested if a lint-error would fail the build here: https://travis-ci.org/github/agenda/agenda/builds/754085260

leonardlin (approved):

koresar (approved):

Issue #1191, opened 2021-01-16
Error: Cannot find module './lib/agenda' in Agenda v4.0.0


marius4896:
Hello,
After updating to 4.0.0 i get an error even when importing the lib 
```
node src/app.js 
\internal/modules/cjs/loader.js:797
    throw err;
    ^

Error: Cannot find module './lib/agenda'
Require stack:
```
Node Version : 12.13.0
NPM Version: 6.14.7

Code ran:

```
require("dotenv").config();
const Agenda = require("agenda");
const agenda = new Agenda({ db: { address: process.env.MONGODB } });
```

Hope this helps
Cheers

simison:
Yep, we'll need new release to fix it. Apologies!

It's about us being on Typescript and needing to use dist instead of sources directly.

simison:
Fix in progress at https://github.com/agenda/agenda/pull/1193

simison:
Hi @marius4896; 4.0.1 fixes the issue! Thanks for patience again. :-)

PR #1193, opened 2021-01-16
Use dist instead of lib and compile TS before publish

simison:
An attempt to fix the build for NPM releases after TypeScript rewrite. (Rewrite PR was https://github.com/agenda/agenda/pull/1123)

Except that I still get `TypeError: Agenda is not a constructor`. 😞  I'm pretty novice with TypeScript so not entirely sure why that's happening.

Feel free to take the PR over!

Fixes https://github.com/agenda/agenda/issues/1191

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,3 @@
-const Agenda = require('./lib/agenda');
+const Agenda = require('./dist');
 
 module.exports = Agenda;
--- a/index.js
+++ b/index.js
@@ -1,3 +1,3 @@
-const Agenda = require('./dist');
+const { Agenda } = require('./dist/agenda');
 
 module.exports = Agenda;
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "agenda",
   "version": "4.0.0",
   "description": "Light weight job scheduler for Node.js",
-  "main": "dist/index.js",
+  "main": "index.js",
   "files": [
     "dist"
   ],
--- a/package.json
+++ b/package.json
@@ -2,14 +2,15 @@
   "name": "agenda",
   "version": "4.0.0",
   "description": "Light weight job scheduler for Node.js",
-  "main": "index.js",
+  "main": "dist/index.js",
   "files": [
-    "lib"
+    "dist"
   ],
   "engines": {
     "node": ">=10.0.0"
   },
   "scripts": {
+    "prepublishOnly": "npm run build",
     "build": "tsc",
     "pretest": "npm run build",
     "test": "npm run mocha",


package.json
```diff
@@ -2,14 +2,15 @@
...
+  "main": "dist/index.js",
   "files": [
-    "lib"
+    "dist"
   ],
   "engines": {
     "node": ">=10.0.0"
   },
   "scripts": {
+    "prepublishOnly": "npm run build",
```
simison, line 13: https://docs.npmjs.com/cli/v6/using-npm/scripts#life-cycle-scripts

package.json
```diff
@@ -2,14 +2,15 @@
   "name": "agenda",
   "version": "4.0.0",
   "description": "Light weight job scheduler for Node.js",
-  "main": "index.js",
+  "main": "dist/index.js",
   "files": [
```
simllll, line 6: we could add 
  "types": "dist/index.d.ts",
here.. (not sure if we have types with the curernt build config?, but if we have, adding this would give the user a lot of beneftis when using this library :))

simllll (approved):
LGTM

leonardlin:
I can't reproduce the problem locally using 'npm link'
I can build my project that is using agenda against agenda-branch.

I can reproduce the problem, if I use the agenda@4.0.0 package in my project.

I compared the package.json against my other npm packages. It looks good.
I would say, publish to 4.0.0 and I can retry again.

package.json
leonardlin: Agree, after fixing the release, we should add types-file with the xports.

leonardlin:
Is a bump to 4.0.1 necessary? Or can our build overwrite the existing 4.0.0 package?

simllll:
> I can't reproduce the problem locally using 'npm link'
> I can build my project that is using agenda against agenda-branch.
> 
> I can reproduce the problem, if I use the agenda@4.0.0 package in my project.
> 
> I compared the package.json against my other npm packages. It looks good.
> I would say, publish to 4.0.0 and I can retry again.

I think something like this is happening here:
Locally all the source files (including tsconfig and other stuff) do exists, therefore when you start your application e.g .with ts-node (typescript is at least somehow needed), you can work with it. But when deployed to npm, tsconfig and other files are stripped away, and only the "ts" files (which is wrong) are deployed in this case. therefore everytime when you try to use it, "node" does not understand what you are trying to do, as it cannot read .ts  files.

simllll:
> Is a bump to 4.0.1 necessary? Or can our build overwrite the existing 4.0.0 package?

npm does not allow overwriting packages (anymore), as far as i know.

simison:
4.0.1 release will be necessary indeed. 👍

simison:
@leonardlin @simllll Can you replicate the issue with `Agenda is not a constructor` using this demo: https://github.com/agenda/agenda-demo

package.json
simison: Yep yep, a good follow up task. 👍

simllll:
> @leonardlin @simllll Can you replicate the issue with `Agenda is not a constructor` using this demo: https://github.com/agenda/agenda-demo or is it just me?
> 
> In case this matters:
> 
> ```
> npx tsc --version
> Version 4.1.3
> ```

It's due to the "import" in index.js of Agenda..
change the line from:
`const Agenda = require('./dist/agenda');`
to
`const { Agenda } = require('./dist/agenda');`

and things should work :)

index.js
```diff
@@ -1,3 +1,3 @@
-const Agenda = require('./lib/agenda');
+const Agenda = require('./dist');
```
simllll, line 1: const { Agenda } = require('./dist/agenda');

package.json
```diff
@@ -2,14 +2,15 @@
   "name": "agenda",
   "version": "4.0.0",
   "description": "Light weight job scheduler for Node.js",
-  "main": "index.js",
+  "main": "dist/index.js",
```
simllll, line 5: revert to "index.js"

simllll (commented):
these two changes are fixing the TypeError: Agenda is not a constructor

simison:
Thanks @simllll, much appreciated!

PS. Exporting types PR welcome. 😉

package.json
simison: Done https://github.com/agenda/agenda/pull/1193/commits/4024ad2f7c37025fc366a1191c1f6ee5d8ac8fe3

index.js
simison: Done https://github.com/agenda/agenda/pull/1193/commits/4024ad2f7c37025fc366a1191c1f6ee5d8ac8fe3

simison:
Released this in 4.0.1

PR #1195, opened 2021-01-16
fix: export declarations

simllll:
this would export the types for agenda and therefore give the whole typescript rewrite a legit reason :)

but unfortunately the types are not ready yet for production:
e.g. agenda.define is currently defined like this:
`export const define = function(this: Agenda, name: string, options: any, processor: () => void) {`
a lot of implementations are just like this:
`agenda.define('example', (job, done) => {`

the current type definition does not allow calling agenda.define with just two parameters, 1. options is any and 2. processor is a required additonal attribute. 

There is function overloading available in typescript, see for example in my fork: https://github.com/hokify/agenda/blob/master/src/index.ts#L275
which need to be implemented here too to get things working correctly. Unsure if function overloading is working though with the current file setup, as it requires that everything is part of a class. Right now all methods are split into separate files,.. could work, but haven't tested it. 

Anyways, I'll open this a draft, anyone is welcome to contribute if someone finds time to transfer the types from https://github.com/hokify/agenda to this one and figures out if function overloading works with the current setup? 

Thanks :)

### DIFF
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,3 +1,3 @@
 import { Agenda } from './agenda';
-export { Agenda }
+export { Agenda };
 export default Agenda;
--- a/lib/index.ts
+++ b/lib/index.ts
@@ -1,1 +1,3 @@
-export { Agenda } from './agenda';
+import { Agenda } from './agenda';
+export { Agenda }
+export default Agenda;
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "4.0.1",
   "description": "Light weight job scheduler for Node.js",
   "main": "index.js",
+  "types": "dist/index.d.ts",
   "files": [
     "dist"
   ],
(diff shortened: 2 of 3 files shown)


simllll:
As a followup, as soon as we have the correct types, we could start changing the tests also to typescript. This helps finding wrong type defs and further improves the code as the tests will benefit from typescript types too :)

leonardlin:
I added types, this 

> 
> 
> this would export the types for agenda and therefore give the whole typescript rewrite a legit reason :)
> 
> but unfortunately the types are not ready yet for production:
> e.g. agenda.define is currently defined like this:
> `export const define = function(this: Agenda, name: string, options: any, processor: () => void) {`
> a lot of implementations are just like this:
> `agenda.define('example', (job, done) => {`
> 
> the current type definition does not allow calling agenda.define with just two parameters, 1. options is any and 2. processor is a required additonal attribute.
> 
> There is function overloading available in typescript, see for example in my fork: https://github.com/hokify/agenda/blob/master/src/index.ts#L275
> which need to be implemented here too to get things working correctly. Unsure if function overloading is working though with the current file setup, as it requires that everything is part of a class. Right now all methods are split into separate files,.. could work, but haven't tested it.
> 
> Anyways, I'll open this a draft, anyone is welcome to contribute if someone finds time to transfer the types from https://github.com/hokify/agenda to this one and figures out if function overloading works with the current setup?
> 
> Thanks :)

I added types: https://github.com/agenda/agenda/pull/1202
I think this change should make sense now.

simllll:
this PR is obsolete

Issue #1187, opened 2021-01-08
Update  default useUnifiedTopology option to mongoDb default


pappyJ:
Displays a warning message in the console if not explicitly declared which is not specified in the project documentation, I fixed it using. 

`agenda.database(DATABASE_LOCAL, 'backups', { useUnifiedTopology: true });` 

but i needed to locate where the code was written in the package to detect it . which might not be easy for all devs using the module

leonardlin:
solved in PR #1200

PR #1200, opened 2021-01-18
1187 use unified topology default

leonardlin:
Based on issue https://github.com/agenda/agenda/issues/1187
It has been bugging me for a while too.

Mongo is changing the default connection mode.
My switching to the new mode now, we can detect issues and be ready when the default-mode switches in the driver.

### DIFF
--- a/lib/agenda/cancel.ts
+++ b/lib/agenda/cancel.ts
@@ -1,5 +1,6 @@
 import { Agenda } from '.';
 import createDebugger from 'debug';
+import { FilterQuery } from 'mongodb';
 
 const debug = createDebugger('agenda:cancel');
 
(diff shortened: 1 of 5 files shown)


simllll (approved):
Lgtm, this is long overdue in my opinion ;-)

lib/agenda/cancel.ts
```diff
@@ -1,16 +1,17 @@
...
+import { FilterQuery } from 'mongodb';
 
 const debug = createDebugger('agenda:cancel');
 
 /**
  * Cancels any jobs matching the passed MongoDB query, and removes them from the database.
  * @name Agenda#cancel
  * @function
- * @param {Object} query MongoDB query to use when cancelling
+ * @param FilterQuery<any> query MongoDB query to use when cancelling
```
simllll, line 11: Not sure if this is needed, in general when all types are in place, a documentation can be generated directly from the typings. No need for extra markups. But it will not harm too, therefore just wanted to point it out ;-)

leonardlin: oh, I didn't know that.
So the generation will derive the type correctly?

That change actually was not supposed to in this PR.
I am working on a type-enhancement PR.

But thanks for letting me know.

simllll: Yop :) which also helps a lot to keep documentation up2date :) i have used this approach in the hokfiy fork, e.g. This https://hokify.github.io/agenda/agenda/4.x/index.html documentation uses all the typescript types

simison:
Remember to update change-log with this. ;-)

lib/agenda/cancel.ts
simison: [Docs](https://agenda.github.io/agenda/) are currently generated based on these, but we can of course change that, or remove those docs entirely.

PR #1202, opened 2021-01-19
Add types

leonardlin:
Added types to most places


lib/agenda/db-init.ts
```diff
@@ -11,9 +11,9 @@ const debug = createDebugger('agenda:db_init');
  * @param collection name or undefined for default 'agendaJobs'
  * @param cb called when the db is initialized
  */
-export const dbInit = function(this: Agenda, collection: string, cb?: (error: Error, collection: Collection<any> | null) => void) {
+export const dbInit = function(this: Agenda, collection: string | undefined, cb?: (error: Error, collection: Collection<any> | null) => void) {
   debug('init database collection using name [%s]', collection);
-  this._collection = this._mdb.collection(collection || 'agendaJobs');
+  this._collection = this._mdb.collection(collection ?? 'agendaJobs');
```
simllll, line 16: this changes behaviour..
```
false || 'agendaJobs' results to "agendaJobs"
false ?? 'agendaJobs' results to false
```
is this intended?

lib/agenda/define.ts
```diff
@@ -12,18 +27,18 @@ const debug = createDebugger('agenda:define');
  * @param options options for job to run
  * @param processor function to be called to run actual job
  */
-export const define = function(this: Agenda, name: string, options: any, processor: () => void) {
-  if (!processor) {
-    processor = options;
+export const define = function(this: Agenda, name: string, options: DefineOptions | (() => void), processor?: () => void) {
```
simllll, line 30: processor is not exactly () => void it has two valid signatures:
`processor: ((job: Job) => Promise<void>) | ((job: Job, done) => void),`

I would suggest you make a `type Processor = processor: ((job: Job) => Promise<void>) | ((job: Job, done) => void);` and use this here twice.

lib/agenda/define.ts
```diff
@@ -12,18 +27,18 @@ const debug = createDebugger('agenda:define');
...
+  if (processor === undefined) {
+    processor = options as (() => void);
```
simllll, line 32: `"as Processor"`

simllll: you could ppbl also improve the check.. e.g. by checking `typeof options === 'function'` ..

lib/agenda/every.ts
```diff
@@ -14,7 +15,7 @@ const debug = createDebugger('agenda:every');
  * @param options - options to run job for
  * @returns Job/s created. Resolves when schedule fails or passes
  */
-export const every = async function(this: Agenda, interval: string | number, names: string | string[], data: any, options: any): Promise<any> {
+export const every = async function(this: Agenda, interval: string, names: string | string[], data: any, options: JobOptions): Promise<any> {
```
simllll, line 18: if i remember right, the interval can also be a number

lib/agenda/every.ts
```diff
@@ -39,7 +40,7 @@ export const every = async function(this: Agenda, interval: string | number, nam
    * @param options options to run job for
    * @return array of jobs created
    */
-  const createJobs = async(interval: string | number, names: string[], data: any, options: any): Promise<Job[] | undefined> => {
+  const createJobs = async(interval: string, names: string[], data: any, options: JobOptions): Promise<Job[] | undefined> => {
```
simllll, line 43: instead of "any" you could use "unknown" .. which forces you to do a type check  in case you want to access a property, which is way more safe.

lib/agenda/index.ts
```diff
@@ -28,6 +28,23 @@ import { stop } from './stop';
 import { findAndLockNextJob } from './find-and-lock-next-job';
 import { Job } from '../job';
 
+export interface AgendaConfig {
```
simllll, line 31: looks good, but you could also "split" it up like I did in the hokify fork. E.g. right now it doesn't make sense if a use provides "db" and "mongo".. by using a construct like this https://github.com/hokify/agenda/blob/c5e21e98b92ef1ba77dab36f68a331c4b493b382/src/index.ts#L85 you could achive a higher level of accuracy

see also https://github.com/hokify/agenda/blob/master/src/types/DbOptions.ts

lib/agenda/index.ts
```diff
@@ -99,21 +116,21 @@ class Agenda extends EventEmitter {
   start!: typeof start;
   stop!: typeof stop;
 
-  constructor(config: any = {}, cb?: (error: Error, collection: Collection<any> | null) => void) {
+  constructor(config: AgendaConfig = {}, cb?: (error: Error, collection: Collection<any> | null) => void) {
```
simllll, line 119: this was also something I was experimenting on.. as right now "from outside" it looks liek there ar eno default values, and you have to check the documetation of agenda again to find out.
another way of approach this , is to set default values in the constrcutor, check out https://github.com/hokify/agenda/blob/c5e21e98b92ef1ba77dab36f68a331c4b493b382/src/index.ts#L75 to see how I did it

lib/agenda/index.ts
```diff
@@ -99,21 +116,21 @@ class Agenda extends EventEmitter {
...
     super();
 
     this._name = config.name;
     this._processEvery = (humanInterval(config.processEvery) ?? humanInterval('5 seconds')) as number; // eslint-disable-line @typescript-eslint/non-nullable-type-assertion-style
-    this._defaultConcurrency = config.defaultConcurrency || 5;
-    this._maxConcurrency = config.maxConcurrency || 20;
-    this._defaultLockLimit = config.defaultLockLimit || 0;
-    this._lockLimit = config.lockLimit || 0;
+    this._defaultConcurrency = config.defaultConcurrency ?? 5;
```
simllll, line 124: again here you change the behaviour slightly by using ?? instead of || ...intended? does it make sense to set a default concurrency of 0?

lib/agenda/index.ts
```diff
@@ -124,8 +141,8 @@ class Agenda extends EventEmitter {
     });
 
     if (config.mongo) {
-      this.mongo(config.mongo, config.db ? config.db.collection : undefined, cb);
-      if (config.mongo.s && config.mongo.topology && config.mongo.topology.s) {
+      this.mongo(config.mongo, config.db ? config.db.collection : undefined, cb); // @ts-expect-error // the documentation shows it should be correct: http://mongodb.github.io/node-mongodb-native/3.6/api/Db.html
```
simllll, line 144: what's the error you get here?

lib/agenda/job-processing-queue.ts
```diff
@@ -47,8 +47,8 @@ JobProcessingQueue.prototype.push = function(this: JobProcessingQueue, job: Job)
  */
 JobProcessingQueue.prototype.insert = function(this: JobProcessingQueue, job: Job) {
   const matchIndex = this._queue.findIndex(element => {
-    if (element.attrs.nextRunAt.getTime() <= job.attrs.nextRunAt.getTime()) {
-      if (element.attrs.nextRunAt.getTime() === job.attrs.nextRunAt.getTime()) {
+    if (element.attrs.nextRunAt!.getTime() <= job.attrs.nextRunAt!.getTime()) {
```
simllll, line 50: nextRunAt is actually not always set, at least if i remember correctly. (never happened to me in production though), so good to go for me, but if we want to be "100%" safe, we should add an additional check like:
`if (element.attrs.nextRunAt && element.attrs.nextRunAt.getTime() <= job.attrs.nextRunAt!.getTime()) {`

lib/agenda/mongo.ts
```diff
@@ -9,7 +9,7 @@ import { Agenda } from '.';
  * @param collection name collection we want to use ('agendaJobs')
  * @param cb called when MongoDB connection fails or passes
  */
-export const mongo = function(this: Agenda, mdb: Db, collection: string, cb?: (error: Error, collection: Collection<any> | null) => void): Agenda {
+export const mongo = function(this: Agenda, mdb: Db, collection: string | undefined, cb?: (error: Error, collection: Collection<any> | null) => void): Agenda {
```
simllll, line 12: why not making collection optoinal? .. `collection?: string`

lib/job/compute-next-run-at.ts
```diff
@@ -17,7 +17,7 @@ export const computeNextRunAt = function(this: Job) {
   const interval = this.attrs.repeatInterval;
   const timezone = this.attrs.repeatTimezone;
   const { repeatAt } = this.attrs;
-  const previousNextRunAt = this.attrs.nextRunAt || new Date();
+  const previousNextRunAt = this.attrs.nextRunAt ?? new Date();
   this.attrs.nextRunAt = undefined;
 
   const dateForTimezone = (date: any): moment.Moment => {
```
simllll, line 23: date is Date isn't it?

lib/job/compute-next-run-at.ts
```diff
@@ -35,7 +35,7 @@ export const computeNextRunAt = function(this: Job) {
   const computeFromInterval = () => {
     debug('[%s:%s] computing next run via interval [%s]', this.attrs.name, this.attrs._id, interval);
     const dateNow = new Date();
-    let lastRun: Date = this.attrs.lastRunAt || dateNow;
+    let lastRun: Date = this.attrs.lastRunAt ?? dateNow;
     let { startDate, endDate, skipDays } = this.attrs;
     lastRun = dateForTimezone(lastRun).toDate();
     const cronOptions: any = { currentDate: lastRun };
```
simllll, line 41: `const cronOptions = { currentDate: lastRun, tz: timezone };` wouuld infer the correct type all time

lib/job/index.ts
```diff
@@ -67,8 +93,10 @@ class Job {
     this.attrs = {
       ...attrs,
       // NOTE: What is the difference between 'once' here and 'single' in agenda/index.js?
-      type: type || 'once',
-      nextRunAt: nextRunAt || new Date()
+      name: attrs.name ?? '',
+      priority: attrs.priority,
+      type: type ?? 'once',
+      nextRunAt: nextRunAt ?? new Date()
```
simllll, line 99: ?? instead of || intended?

lib/job/remove.ts
```diff
@@ -5,6 +5,6 @@ import { Job } from '.';
  * @name Job#remove
  * @function
  */
-export const remove = function(this: Job) {
+export const remove = async function(this: Job) {
```
simllll, line 8: no need for async or? instead you could add a `:Promise<..>` as return value

lib/job/save.ts
```diff
@@ -6,6 +6,6 @@ import { Job } from '.';
  * @function
  * @returns instance of Job resolved after job is saved or errors
  */
-export const save = function(this: Job) {
+export const save = async function(this: Job) {
```
simllll, line 9: no async, but add return type

lib/utils/process-jobs.ts
```diff
@@ -227,7 +227,7 @@ export const processJobs = async function(this: Agenda, extraJob: Job) {
 
     // If the 'nextRunAt' time is older than the current time, run the job
     // Otherwise, setTimeout that gets called at the time of 'nextRunAt'
-    if (job.attrs.nextRunAt <= now) {
+    if (!job.attrs.nextRunAt || job.attrs.nextRunAt <= now) {
```
simllll, line 230: not quite sure about this... shouldn't it be 
    `if (job.attrs.nextRunAt && job.attrs.nextRunAt <= now) {
`
to keep the logic the same? "undefined < Date" always returns false

lib/utils/process-jobs.ts
```diff
@@ -253,7 +253,7 @@ export const processJobs = async function(this: Agenda, extraJob: Job) {
           // This means a job has "expired", as in it has not been "touched" within the lockoutTime
           // Remove from local lock
           // NOTE: Shouldn't we update the 'lockedAt' value in MongoDB so it can be picked up on restart?
-          if (job.attrs.lockedAt < lockDeadline) {
+          if (!job.attrs.lockedAt || job.attrs.lockedAt < lockDeadline) {
```
simllll, line 256: same for this check, shouldn it be:
`          if (job.attrs.lockedAt && job.attrs.lockedAt < lockDeadline) {
`

simllll (commented):
nice work :-) I just added some comments, most of them are just things that I remembered from refactoring it last time, most of it are nothing to worry about it though :)

lib/agenda/db-init.ts
leonardlin: hmm, yeah I was thinking the same thing. I was aware that they don't behave the same.
I was thinking that back in the days people used "||" but to do more precisely "??"

So while changing I checked that all the unit-tests still run fine.

I was hoping to cleanup null vs undefined un-preciseness. 
By introducing this difference, I thought that going forward we are more aware when we want to use "undefined" and when we actually want to use "null".

But I think will go through the places and see, where mixing null/undefined will cause issues and where we use '??' to express default-options.

lib/job/index.ts
leonardlin: it's intended here, as we check for "undefined" not "null".
I will check type definitions that 'null' is invalid.

lib/job/remove.ts
leonardlin: Typescript/eslint enforces that that "async" - "Promise" come in tandem.
You can't have one without the other.

lib/utils/process-jobs.ts
leonardlin: There is an "!" at the beginning.

To maintain the old logic the if - condition is.

(job.attrs.nextRunAt == falsly) OR (<= now)

lib/agenda/every.ts
leonardlin: I just checked again. "interval" is passed down further down the chain and eventually ends up in JobAttribute.repeatInterval.
This fields needs to be a "human readable" time-interval or cron-expression.

lib/agenda/every.ts
leonardlin: changed to unknown

leonardlin:
@simllll  thanks for the review comment. I updated the PR

lib/agenda/index.ts
koresar: I'd second that.

koresar (changes requested):
I am by no means a TS expert, but I have only one question really.
I'd love this merged ASAP.

In fact, I merged latest `master` to this branch and then fixed all the hiccups. FYI @leonardlin

leonardlin:
@koresar  thanks for updating the branch. I'm eager to merge it as well but was waiting for an approver.

About "??" vs "||"

In places where we were unprecise (legacy), I left it "||"

Newer places where we want to capture "default value" I used "??"

Example case:
```
config.defaultLockLimit = 0;
this._defaultLockLimit = config.defaultLockLimit ?? 5;
// this._defaultLockLimit == 0
```
vs
```
this._defaultLockLimit = config.defaultLockLimit || 5;
// this._defaultLockLimit == 5
```

What should it be?

koresar:
If we put ?? it would change the default, that would be a breaking change.
If the next release is v5 then it's okay. But it's rather do fewer breaking
changes in projects like this.

leonardlin:
good point. I will check where I changed it and reverse it.

leonardlin:
@koresar  I updated the ?? || thing. can you review

koresar (approved):

lib/agenda/define.ts
OmgImAlexis: `done` should be typed.

PR #1211, opened 2021-01-27
Format using Prettier.

harisvsulaiman:
### Use prettier to format agenda.
Based on the discussion at https://github.com/agenda/agenda/discussions/1209.

Note: Linting fails for now. I wanted to know what rules should be enforced on the code or should it be the prettier defaults?
cc @simison @simllll @OmgImAlexis


koresar:
The most liveable mounting rules I know - the default built-in ESLint rules. We would not need to install, maintain additional dependencies.

harisvsulaiman:
Now I am using eslint defaults along with prettier defaults.
The pre-commit hook via lint staged checks for issues

package.json
```diff
@@ -57,75 +57,37 @@
...
     "delay": "4.4.0",
     "eslint": "7.18.0",
-    "eslint-config-xo": "0.34.0",
-    "eslint-config-xo-typescript": "0.37.0",
+    "eslint-config-prettier": "^7.2.0",
     "eslint-plugin-ava": "11.0.0",
     "eslint-plugin-eslint-comments": "3.2.0",
     "eslint-plugin-import": "2.22.1",
     "eslint-plugin-node": "11.1.0",
     "eslint-plugin-unicorn": "24.0.0",
```
simison, line 71: `eslint-plugin-unicorn` can be removed?

harisvsulaiman: Yes. You have to remove the eslint warning inside a file.

koresar (changes requested):
.eslintcache needs to be removed and added to gitignore.

.eslintrc
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": ["plugin:@typescript-eslint/recommended", "prettier"]
```
koresar, line 2: Can we move this file contents to packege.json? Similar to other configurations.

package.json
koresar: If we are using eslint defaults, we can pretty much remove most of the things which start with "eslint-".
Am I right?

harisvsulaiman: Yes. But eslint default are strict on "any" types.
I changed the error to a warning.
We can fix that after we merge this

.eslintrc
harisvsulaiman: I think the editor plugin for eslint picks up the config easily this way.what do you think @simison ?

harisvsulaiman (commented):
Add warning for some defaults now.

.eslintrc
koresar: Editors pick up whatever config. Either in package.json, or in .eslintrc.

harisvsulaiman: Ok.

koresar (approved):

koresar:
I ran the master on my machine, then checked out the PR branch and tested it.
LGTM!
Thanks @harisvsulaiman

Issue #1224, opened 2021-02-01
Exception during upgrade to agenda v4.0.1 from v2.1.0 


talik077:
I was trying recently to upgrade agenda to the  current latest version. I was going through the changelog to make sure I'm not missing any breaking changes. However, after the upgrade I've noticed that my jobs fail `failed to calculate nextRunAt due to invalid repeat interval` error. **I was able to fix it so I hope these steps will help anyone who is trying to upgrade to agenda v4 from the older version.** also I think it's worth updating the breaking changes list and fix it. 

### The steps that I took
According to the changelog, v4 switched from ncb000gt/node-cron to harrisiirak/cron-parser. 
since the `repeatInterval` was already in the correct cron-job format, meaning that I didn't need to update the month part in my cron-job expression, it was easy to point that the issue is not due to parsing exception. 
Only when I debug agenda and I was able to figure out the real error: `Cannot read property 'toISOString' of null` because `nextRunAt` is null. the only place that could override this value is: https://github.com/agenda/agenda/blob/234fff036e46bb8d70f34177e79a843e71a327ea/lib/job/compute-next-run-at.ts#L86-L93  
As you can see, https://github.com/agenda/agenda/pull/361 introduced new parameters: `startDate` and `endDate`. those parameters were missing from my old records, meaning they were undefined. but the condition [here](https://github.com/agenda/agenda/blob/234fff036e46bb8d70f34177e79a843e71a327ea/lib/job/compute-next-run-at.ts#L86) checks whether they are null and this introduces a bug. 

**In order to fix the issue, I had to update all current records and set `startDate` and `endDate` with default null values.** 

see the screenshot below for comparing how v2 vs v4 document schemas look like (old vs new)
### OLD
![image](https://user-images.githubusercontent.com/17844958/106453674-55a7df80-6492-11eb-8375-762c8180a435.png)
### NEW
![image](https://user-images.githubusercontent.com/17844958/106453707-622c3800-6492-11eb-8244-7816d84821e5.png)

We can also see that v4 automatically sets null `startDate` and `endDate` by default on any new document.

PR #1227, opened 2021-02-02
Check endDate and startDate are defined when calculating nextRunAt

talik077:
Fixing the issue with https://github.com/agenda/agenda/issues/1224
For backward compatibility check that endDate and startDate are defined when calculating `nextRunAt` value.
When upgrading from an older version of agenda those fields could be missing from the job and fail the computation of `nextRunAt` with 'failed to calculate nextRunAt due to an invalid repeat interval' exception.

image


harisvsulaiman (approved):

leonardlin:
@harisvsulaiman was there a reason why you didn't merge it? is there something open?

harisvsulaiman:
@leonardlin forgot to merge it after review😄

PR #1236, opened 2021-02-04
add npm cache to ci

leonardlin:


harisvsulaiman (approved):

harisvsulaiman:
@leonardlin How fast is the ci after cache was enabled?

leonardlin:
On out project the impact is not very high. Like 10sec.

For projects with more dependencies and more sequential builds it saves a lot.

PR #1243, opened 2021-02-08
chore: emit type declarations

boredland:
is there any particular reason this isn't being done yet? In my eyes big parts of the definitelyTyped-pkg are outdated, while we could easily bundle the types.

### DIFF
--- a/lib/agenda/define.ts
+++ b/lib/agenda/define.ts
@@ -13,9 +13,26 @@ export enum JobPriority {
 }
 
 export interface DefineOptions {
+  /**
+   * Maximum number of that job that can be running at once (per instance of agenda)
+   */
   concurrency?: number;
+
+  /**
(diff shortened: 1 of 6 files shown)


leonardlin:
Thanks @boredland  our types were not really up to date until recently.

@simllll this is parallel to your PR: https://github.com/agenda/agenda/pull/1195
Are you ok if we accept this PR #1243 it seems good to go for me.

After this PR we can let the guys maintaining @types/ know that their package is not necessary anymore

leonardlin (approved):
Looks good to me. Thanks so much

boredland:
> After this PR we can let the guys maintaining @types/ know that their package is not necessary anymore

I could try to take take of that. At least you should mention in the release notes, that it got deprecated.

leonardlin:
@boredland thanks so much for this. Thanks for reaching out to @types/agenda team about this.

PR #1261, opened 2021-03-01
Fix Bree typo in README

jhult:


leonardlin (approved):

niftylettuce:
This is not a typo, see https://github.com/agenda/agenda/commit/c98ba5c7c4f7f2ce4f8efe822138618232861f8e

We would need to add another column for Bree, or drop Bee in favor of Bree and update the column values.

Bree itself does not use Redis nor Mongo, it's a DIY system.

manansharma1991:
@niftylettuce Sorry for being harsh. But bree just sucks coz you want to appear everywhere and showcase that you have created a self claimed "THE Best job scheduler". Whereas when someone asks you, you decide to delete their issues.

CleanShot 2021-11-13 at 09 22 11@2x

PR #1263, opened 2021-03-01
Ask TS to emit ES2017 instead of ES5.

koresar:
We do not support Node lower than 10. However, I know people using Agenda v4 with Node 8. So, let's not do too much of the breaking changes here by using ES2017 (Node 8) instead of the more logical ES2018 (Node 10).


koresar:
Superseeded by #1268

Issue #1267, opened 2021-03-04
Types are incorrectly exported


RA80533:
```ts
import { Agenda } from "agenda";
```

```console
$ yarn run tsc
Could not find a declaration file for module 'agenda'. '/[…]/node_modules/agenda/index.js' implicitly has an 'any' type.
  Try `npm i --save-dev @types/agenda` if it exists or add a new declaration (.d.ts) file containing `declare module 'agenda';`ts(7016)
```

While [`agenda@4.1.0`]() is known to have problems relating to this (see #1265 and #1266), they still exist in the most recent commit on the `master` branch, adc22b8 (as of the time of posting).

Reverting the change made in 4024ad2 to package.json would solve the issue as it currently exists. The driving factor at play here is that the root index.js is considered the entry point of the module but is never part of the TypeScript compilation process; however, if it was processed by TypeScript, compilation would fail because it refers to files resulting from the compilation process.

### Solutions

1. Update the 'main' property in package.json to "./dist/index.js" and delete the root index.js file
2. Create a type declaration file for the root index.js file by hand and update the package.json 'types' property accordingly

/cc @simllll

simllll:
Hi @RA80533 , i will quickly look into it, but I believe I had this issue already and I considered it here once already https://github.com/agenda/agenda/pull/1195/files
I'll check and will create a PR :)

RA80533:
Yeah, it seemed like it was something considered looking back at https://github.com/agenda/agenda/pull/1193#discussion_r559038508. Thanks for the quick response.

simllll:
yeah I think there is actually no problem with setting main to dist/index.js directly, as there is also a default export. I will test it quickly

PR #1268, opened 2021-03-04
fix: typescript export needs es6

simllll:
fixes https://github.com/agenda/agenda/issues/1267

There are two issues:
1.) there is only one default export, no named export. People are assuming a named export will work fine too. Which is not direclty a bug, but a nice to have ;)
2.) The export with es5 does not work correclty, as es5 is not aware of class exports. 
Test:

```ts
import { Agenda } from 'agenda';
console.log('AGENDA', Agenda);
```
```ts
import Agenda from 'agenda';
console.log('AGENDA', Agenda);
```
all should output: 
`AGENDA [class Agenda extends EventEmitter]
`
But with es5 they output:
`AGENDA [Function: Agenda]
`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1079,8 +1079,8 @@ Please provide us with as much details as possible such as:
 
 #### To turn on logging, please set your DEBUG env variable like so:
 
-- OSX: `DEBUG="agenda:*" node index.js`
-- Linux: `DEBUG="agenda:*" node index.js`
+- OSX: `DEBUG="agenda:*" ts-node src/index.js`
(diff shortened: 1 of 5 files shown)


RA80533 (approved):
The root index.js file can now be safely removed since it's no longer used as the entry point.

simllll:
> The root index.js file can now be safely removed since it's no longer used as the entry point.

Thanks @RA80533 ,pinging @koresar for final okay :)

koresar (approved):

PR #1272, opened 2021-03-09
Fix schedule for getting a single name

thebestnom:
Sorry if it looks bad, I did it quickly from my phone as an example

### DIFF
--- a/lib/agenda/schedule.ts
+++ b/lib/agenda/schedule.ts
@@ -13,12 +13,14 @@ const debug = createDebugger("agenda:schedule");
  * @param data data to send to job
  * @returns job or jobs created
  */
-export const schedule = function (
+export function schedule(this: Agenda, when: string | Date, names: string, data: any): Promise<Job>;
+export function schedule(this: Agenda, when: string | Date, names: string[], data: any): Promise<Job[]>;
+export function schedule (
   this: Agenda,
   when: string | Date,
-  names: string[],
+  names: string | string[],
   data: any
-): undefined | Promise<Job | Job[]> {
+): Promise<Job | Job[]> {
   /**
    * Internal method that creates a job with given date
    * @param when when the job gets run
@@ -41,7 +43,7 @@ export const schedule = function (
   /**
    * Internal helper method that calls createJob on a names array
    * @param when when the job gets run
-   * @param of jobs to run
+   * @param names names of jobs to run
    * @param data data to send to job
    * @returns jobs that were created
    */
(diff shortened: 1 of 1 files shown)


koresar:
I'm not an expert. But to me, deflating a type to use some sounds like we better be without the type.

I suggest removing both new types. And using string|string[] and Job|Job[]

Also, I'm not a TS expert. So let's just for other onions.

leonardlin:
I agree with the problem, that we have a lot of function definition that take multiple types as argument.

I don't know if it's got practice to create a type as a workaround.

I think it would be better to cleanup the code, but I defer here to others as well.
Whenever I worked on the types, the cleanup was often non-trivial, that I just went with the bad types.

thebestnom:
I dont think it's a workaround, we can easily infer return type based on sent type, it not that uncommon to do (users dont have to insert the generic, it's automatically infered based on argument)

thebestnom:
Other way is to do function overloading like https://stackoverflow.com/questions/57244402/typescript-infer-return-type-from-used-arguments

simllll:
Would recommend function overloading, but for getting this to work I guess it is mandatory to have classes. And right now the code isn't structured like this (more the opposite with all the overloading from different files), but I also would vote for it to get it into a nicely dressed class style ;-) see @hokify/agenda as example

thebestnom:
I don't think function overloading require class, let me test that

simllll:
> I don't think function overloading require class, let me test that

It's possible to just define multiple defintions for overloading, but not with prototype overwrites like it is now. At least as far as I know. Besides that there are some other issues with exporting the type in case it's not bound to a class, but that shouldn't be a deal breaker.

e.g. 
this works:
```
function add(a: number, b: number, c: number): number;
function add(a: number, b: number): any;
function add(a: string, b: string): any;

function add(a: any, b: any, c?: any): any {
```

but exporting will not, or at least I'm not aware of any syntax that would be "usable" enough (you always can overload it manually  via defintion file of course;)).
```
export const database = function (this: Agenda)
export const database = function (
  this: Agenda,
  url?: string
  );
export const database = function (
  this: Agenda,
  url?: number
  );
export const database = function (
  this: Agenda,
  url?: any
  ) { .. }
```

thebestnom:
testing this now

thebestnom:
Ok, pushed with operator overloading, looks like it's working (tested by creating new Agenda and calling schedule

thebestnom:
there is a second way of doing this with const but that requires return type any (no return check) with https://stackoverflow.com/a/53143568

koresar (approved):

koresar:
Beautiful!
@simllll should we merge?

simllll (approved):

PR #1279, opened 2021-03-12
fix: Typescript optional parameters

Igor-lkm:
Problem:

image

Workaround:

```javascript
await agenda.every('10 minutes', 'some job', undefined, undefined);
```

What this MR will do:
- will fix optional parameters in type declarations

```javascript

// instead of this:
await agenda.every('10 minutes', 'some job', undefined, undefined);

// we will be able to do this:
await agenda.every('10 minutes', 'some job');
```


koresar (approved):
Brilliant!

PR #1291, opened 2021-04-04
Fixes #1286 do not try use `null` as object

### LINKED ISSUE #1286: Misleading failure when nextRunAt is null in computeFromInterval
Hi

I'm having an issue with the `computeFromInterval` function in `compute-next-run-at.ts`. When the `nextRunAt` property passes the `endDate`, the function incorrectly marks the job as failed due to some (I believe) mistakes in the function logic and the job is saved into the database with a +1 fail count and a misleading fail reason.

This is my job instance:
```json
{ 
    "_id" : "60549f13e85acb3f7c284c30", 
    "name" : "test-task", 
    "type" : "single", 
    "data" : {
        "username":  "Philip"
    }, 
    "endDate" : "2021-03-25T14:59:00.000+0000", 
    "lastModifiedBy" : "agenda-mongodb", 
    "nextRunAt" : "2021-03-25T14:59:00.000+0000", 
    "priority" : 0, 
    "repeatInterval" : "* * * * *", 
    "repeatTimezone" : null, 
    "skipDays" : null, 
    "startDate" : "2021-03-25T13:40:00.000+0000", 
    "lockedAt" : "2021-03-25T14:58:59.733+0000", 
    "lastRunAt" : "2021-03-25T14:55:00.003+0000", 
    "failCount" : 33, 
    "failReason" : "failed to calculate nextRunAt due to invalid repeat interval", 
    "failedAt" : "2021-03-25T14:55:00.050+0000", 
    "lastFinishedAt" : "2021-03-25T14:55:00.055+0000"
}
```

After a little bit of digging, I figured out that you set the `nextRunAt` property to `null` if `nextDate` passes `endDate` `[line 86]`
```ts
      if (endDate) {
        const endDateDate: Date = moment
          .tz(moment(endDate).format("YYYY-MM-DD"), timezone!)
          .toDate();
        if (nextDate > endDateDate) {
          nextDate = null;
        }
      }

      this.attrs.nextRunAt = nextDate;
```
but right on the next line, you attempt `[line 96]`
```ts
      debug(
        "[%s:%s] nextRunAt set to [%s]",
        this.attrs.name,
        this.attrs._id,
        this.attrs.nextRunAt!.toISOString()
      );
```
which throws an error, because you are calling `toISOString()` on `null`. This causes the process to jump into the `catch` block, where you incorrectly log this `[line 104]`
```ts
      debug(
        "[%s:%s] failed nextRunAt based on interval [%s]",
        this.attrs.name,
        this.attrs._id,
        interval
      );
```
instead of just logging the actual error (it is a debug log after all, it should say exactly what went wrong), and then you attempt to calculate `nextRunAt` again, this time using `humanInterval`. `[line 120]`
```ts
        } else {
          this.attrs.nextRunAt = new Date(
            lastRun.getTime() + (humanInterval(interval) ?? 0)
          );
```
This results in `nextRunAt` being an `Invalid Date {}`. You're probably relying on this, but I believe it is incorrect to be using the `catch` block as pretty much a switch to decide which parser to use. Lastly, in the `finally` block, this whole ordeal is marked as a failure (even though it actually isn't, the job was correctly evaluated to never run again) and `fail` is called with a misleading message. `[line 133]`
```ts
    } finally {
      if (Number.isNaN(this.attrs.nextRunAt!.getTime())) {
        this.attrs.nextRunAt = undefined;
        debug(
          "[%s:%s] failed to calculate nextRunAt due to invalid repeat interval",
          this.attrs.name,
          this.attrs._id
        );
        this.fail(
          "failed to calculate nextRunAt due to invalid repeat interval"
        );
      }
    }
```
The repeat interval is actually valid.

Amusingly enough, incorrectly (IMO) jumping into the catch block actually *saves* the process, because if the catch block logic did not set `nextRunAt` to `Invalid Date`, the `if` clause in the `finally` block would simply cause a fatal error. `[line 134]`
```ts
if (Number.isNaN(this.attrs.nextRunAt!.getTime()))
```
Here `getTime()` is called, once again, on a potentially null property.

koresar:
Detailed explanation of the issue is here: #1286

@Aravill could you please test if this fix works for you? I assume it does not.

### DIFF
--- a/lib/job/compute-next-run-at.ts
+++ b/lib/job/compute-next-run-at.ts
@@ -97,7 +97,7 @@ export const computeNextRunAt = function (this: Job): Job {
         "[%s:%s] nextRunAt set to [%s]",
         this.attrs.name,
         this.attrs._id,
-        this.attrs.nextRunAt!.toISOString()
+        this.attrs.nextRunAt?.toISOString()
       );
       // Either `xo` linter or Node.js 8 stumble on this line if it isn't just ignored
     } catch {
@@ -131,7 +131,7 @@ export const computeNextRunAt = function (this: Job): Job {
         // Either `xo` linter or Node.js 8 stumble on this line if it isn't just ignored
       } catch {}
     } finally {
-      if (Number.isNaN(this.attrs.nextRunAt!.getTime())) {
+      if (!this.attrs.nextRunAt?.getTime()) {
         this.attrs.nextRunAt = undefined;
         debug(
           "[%s:%s] failed to calculate nextRunAt due to invalid repeat interval",
@@ -168,7 +168,7 @@ export const computeNextRunAt = function (this: Job): Job {
         "[%s:%s] nextRunAt set to [%s]",
         this.attrs.name,
         this.attrs._id,
-        this.attrs.nextRunAt!.toISOString()
+        this.attrs.nextRunAt?.toISOString()
       );
     } else {
       this.attrs.nextRunAt = date(repeatAt);
@@ -176,7 +176,7 @@ export const computeNextRunAt = function (this: Job): Job {
         "[%s:%s] nextRunAt set to [%s]",
         this.attrs.name,
         this.attrs._id,
-        this.attrs.nextRunAt!.toISOString()
+        this.attrs.nextRunAt?.toISOString()
       );
     }
   };


simllll (approved):
Lgtm, without testing I guess this should actually cover the issue of aravill.

JakubMifek:
This won't work. The finally block will fail once again since nextRunAt is null. Therefore the if statement will be evaluated to true and the algorithm will incorrectly fail.

If I could recommend, rewrite the whole method and stop using catch for switching between different kinds of evaluation, it's ineffective and not clean at all.. If you do that, you don't need `finally` block and can handle fails where (if) they happen.

simllll:
But now the access to nextrunat is optional chained, therefore i assume it should not fail anymore? In any case, i would say the optional chaining should be merged into master as this is actually more correct in my opinion :)

koresar:
@JakubMifek I like your recommendation very much! Could you please supply a PR which implements that?

JakubMifek:
> 
> 
> But now the access to nextrunat is optional chained, therefore i assume it should not fail anymore? In any case, i would say the optional chaining should be merged into master as this is actually more correct in my opinion :)

It will fail, because:
```js
if (!this.attrs.nextRunAt?.getTime()) { ... this.fail(...); }
```

will be evaluated to:
```js
if (!(null?.getTime())) { ... }
```

which will be evaluated to:
```js
if (!(undefined)) { ... this.fail(...); }
```

which will result in:
```js
if (!false) { ... }
```

therefore:
```js
if (true) { ... }
```

I hope I didn't make mistake but that's how it works in my head 😅 

if i understand the issue of Aravill, he just needs the job to stop, not fail.

> @JakubMifek I like your recommendation very much! Could you please supply a PR which implements that?

I would really like to but I am kinda overwhelmed with my own work at the moment :-/ I am writing here only because I collaborate with Aravill and got a bit invested in the issue..

Aravill:
Hi, I'm currently unable to test the fix, but I agree with what @JakubMifek says. The job did not fail, it was correctly evaluated to not be executed, so it should not be marked as failed, `this.fail()` shouldn't be called. Checking specifically for `null` in the `if` statement of the `finally` block should suffice.

koresar:
Great! We are expecting for your PRs. I'll close this one as soon as you guys submit a better version.

Big thanks!!!

leonardlin (approved):

koresar:
Waited for 7 days. Since this is the only proposed PR so far I'm going to merge it and wait for other (proper) fixes of this issue.

JakubMifek:
This PR does not fix the issue as mentioned before. If anything, issue https://github.com/agenda/agenda/issues/1286 should remain open since it was not fixed yet.

koresar:
@JakubMifek would you submit a PR to fix it properly?

Issue #1266, opened 2021-03-03
typing issue


bwgjoseph:
Hi,

I am trying out `agenda` but facing with types issue hence, unable to proceed. Could be related to #1218, and #1265 

The example usage of initializing `agenda` is `const agenda = new Agenda({ db: { address: mongoConnectionString } });`. However, the typing does not match what is in the docs.

```js
const agenda = new Agenda({
  db: {
    address: 'mongodb://...',
  },
});
```

https://github.com/agenda/agenda/blob/adc22b8b662e6e59364cf6bb4f2518e00606e482/lib/agenda/index.ts#L36-L51

The definition is as such, and since `db.collection` and `db.options` are not declared as optional (?), it will throw type error like such

```
Type '{ address: string; }' is missing the following properties from type '{ address: string; collection: string; options: MongoClientOptions; }': collection, options
```

---

The 2nd issue is the constructor error

```js
TypeError: agenda_1.Agenda is not a constructor
    at Object.<anonymous> (Z:\scheduler-service\src\agenda.ts:4:16)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Module._compile (Z:\scheduler-service\node_modules\source-map-support\source-map-support.js:547:25)
    at Module.m._compile (C:\Users\Joseph\AppData\Local\Temp\ts-node-dev-hook-769800944806712.js:69:33)
    at Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at require.extensions.<computed> (C:\Users\Joseph\AppData\Local\Temp\ts-node-dev-hook-769800944806712.js:71:20)
    at Object.nodeDevHook [as .ts] (Z:\scheduler-service\node_modules\ts-node-dev\lib\hook.js:63:13)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Module.require (internal/modules/cjs/loader.js:952:19)
[ERROR] 01:02:28 TypeError: agenda_1.Agenda is not a constructor
```

Using `agenda: 4.1.0` and `typescript: 4.1.5`

royibernthal:
Joining the issue.

I'll add that the typings of agenda.every and agenda.schedule are incorrect as well.
There are possibly other mistakes.

Need to redo the typings and actually test them :)

BonnieMilian:
Joining the issue too, not able to start the agenda because of that error.
`agenda_1.Agenda is not a constructor`

[...]

PR #1298, opened 2021-04-17
fix: export cjs and es

simllll:
as commented in #1266, this one would restore the old behaviour and introduces a new entrypoint for es modules.

for CJS (javascript):
`const Agenda = require('agenda');`

or for module (typescript):
`import Agenda from 'agenda/es'`

This is a "another" breaking change though, as a named import would not work with the default entrypoint anymore (didn't work before, but did now in the patch releases)


README.md
```diff
@@ -60,6 +60,19 @@ Install via NPM
...
+
+for regular javascript code, just use the default entrypoint
+```js
+const Agenda = require('agenda');
+```
+
+For Typescript, Webpack or other module imports, use `agenda/es` entrypoint:
+e.g.
+```ts
+import { Agenda } from 'agenda/es';
```
leonardlin, line 73: I would make a note that people migrating from @types/agenda should change their imports 
from: import Agenda from 'agenda'
to: import { Agenda} from 'agenda/es'

leonardlin (dismissed):
I'm not an expert on module/exports etc. therefore I yield to other people.
But I added note for people migrating from @types/agenda

README.md
simllll: added it to the readme

leonardlin (dismissed):

thebestnom:
Couldn't you just add 
```ts
export = Agenda
```
In index.ts?

simllll:
> Couldn't you just add
> 
> ```ts
> export = Agenda
> ```
> 
> In index.ts?

Unfortunately not, this would overwrite all other exports and would therefore not work correctly. It would only fix the plain JS includes, but this does my PR too.

koresar (approved):

simllll:
Thanks for the additional fixes and tests! Lgtm too, feel free to merge

PR #1314, opened 2021-05-15
Add Node.js v16 in tests

simison:
Adds Node.js v16 into test rig.

Happy to keep v10 if anyone thinks it's still important. It's no longer maintained. Edit: we're keeping it.

https://nodejs.org/en/about/releases/

Meanwhile, let's start test for new LTS active.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node-version: [12.x, 14.x, 16.x]
+        node-version: [10.x, 12.x, 14.x, 16.x]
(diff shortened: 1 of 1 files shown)


simison:
Related https://github.com/agenda/agenda/pull/1306

koresar:
I'd rather not rush with deprecating things without an actual need. A need could be :

- A feature from node 12 but not in node 10
- Causes compatibility issues with Agenda's prod or dev dependencies
- Etc

Chances are, there are a lot of people using Agenda in node 10 or even 8. Also, having this PR merged would probably mean we have to release another MAJOR. It's too many major releases as to my taste

Let's just add node 16 at this stage.

simison:
Fair enough! Thanks for thinking it through. 🙂

PR #1340, opened 2021-06-16
Modifies JobAttributes.lockedAt to allow null values

tjramage:
Looks like there was a minor oversight in the recent TypeScript rewrite. Without this change, all of our jobs are constantly locked since `lockedAt` is never reset back to `null` again.

### DIFF
--- a/lib/job/index.ts
+++ b/lib/job/index.ts
@@ -54,7 +54,7 @@ export interface JobAttributes<
   /**
    * Date/time the job was locked.
    */
-  lockedAt?: Date;
+  lockedAt?: Date | null;
 
   /**
    * The priority of the job.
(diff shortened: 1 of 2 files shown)


tjramage:
@koresar – any idea on what the review process in with this? 

FYI: I think there is a deeper problem here. See https://github.com/agenda/agenda/issues/1341#issuecomment-863841661 where a similar issue is happening with `nextRunAt` also.

tjramage:
@koresar – Sorry to nudge, but would be great if we could get this moving. Who might we need involved to progress this further?

koresar (approved):

koresar:
Published in v4.2.0.
Thanks for your contribution!

PR #1346, opened 2021-06-27
Update run.ts

liushmh:
if save to db failed, finished should stay false, otherwise, the job could never be completed.


liushmh:
Be more specific:
 if save is failed, the status of 'old' job is 'finished', but field 'lastFinishedAt' is not updated into database,
so is-running.ts will always return true. 
In this case, the real 'new' job will never start.

harisvsulaiman:
@liushmh Please write a test case for the problem you are trying to solve

PR #1368, opened 2021-08-05
Fix define Processor typing

agix:
Processor type used in define doesn't use generic typing.

If you try this:
```
import Agenda, { Job } from 'agenda'

const agenda = new Agenda()

const run = async function(job: Job<{test: string}>) {
  return Promise.resolve()
}

agenda.define('test', run)
```

It would raise the following typescript error :
```
Type '(job: Job<{ test: string; }>) => Promise<void>' is not assignable to type '(job: Job<JobAttributesData>) => Promise<void>'.
```

Job class use generic type but default to `JobAttributesData` if none is provided https://github.com/agix/agenda/blob/73184aaa8309e5e89e0cfab766b2cdd6d80e9605/lib/job/index.ts?plain=1#L127

Which is exactly the case in the current Processor definition : https://github.com/agenda/agenda/blob/master/lib/agenda/define.ts#L39-L41

### DIFF
--- a/lib/agenda/define.ts
+++ b/lib/agenda/define.ts
@@ -36,9 +36,9 @@ export interface DefineOptions {
   priority?: JobPriority;
 }
 
-export type Processor =
-  | ((job: Job) => Promise<void>)
-  | ((job: Job, done: () => void) => void);
+export type Processor<T> =
+  | ((job: Job<T>) => Promise<void>)
+  | ((job: Job<T>, done: () => void) => void);
 
 /**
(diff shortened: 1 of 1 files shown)


harisvsulaiman:
@agix We can merge this after #1358 Where this is an explicit issue

agix:
Ok thank you.

imsergiobernal:
JobAttributesData should not be generic typed because is an external resource subject to async manipulations. It should be unknown always.

PR #1372, opened 2021-08-09
use Transactions for agenda

harisvsulaiman:
Signed-off-by: Haris Sulaiman <harisvsulaiman@gmail.com>

### DIFF
--- a/lib/agenda/create.ts
+++ b/lib/agenda/create.ts
@@ -1,6 +1,7 @@
 import createDebugger from "debug";
 import { Job } from "../job";
 import { Agenda } from ".";
+import { ClientSession } from "mongodb";
 
 const debug = createDebugger("agenda:create");
 
@@ -11,7 +12,12 @@ const debug = createDebugger("agenda:create");
  * @param name name of job
  * @param data data to set for job
  */
-export const create = function (this: Agenda, name: string, data: any): Job {
+export const create = function (
+  this: Agenda,
+  name: string,
+  data: any,
+  session?: ClientSession
+): Job {
   debug("Agenda.create(%s, [Object])", name);
   const priority = this._definitions[name]
     ? this._definitions[name].priority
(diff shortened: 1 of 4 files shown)


harisvsulaiman:
I would appreciate it if someone could help me with tests.

koresar:
I took a look at the changes. They are great!
Sorry, I can't help with this. Also, looks like you're on your own.
You can try asking people in Agenda Slack maybe?

harisvsulaiman:
@koresar Sure I will post it there. I will take my time completing this pr then.

harisvsulaiman:
@pdfowler Thank you for the feedback. 

> I would vote no support for 4.0, but see some options for handling it, from simplest to more involved: document that sessions require 4.0 (minimum); check DB version on connect, throw error if enabled and <4; Log warning and only enable sessions if >= 4.0.

We can check if sessions are supported when we initialize agenda. Then disable it altogether.

> My gut is to leave that to a separate PR but I don't have sufficient context to say one way or the other

I think we should make that optional too.

I haven’t gotten around to writing tests for this implementation, I would appreciate it if you could lend a hand.

simllll:
Hi all, I was looking into this out of curiosity. Am I right that the main use case for this is that a job's database queries are then handled by one "transaction". First I was thinking that this is a very interesting idea, but I don't really get it now:
* Would that mean agenda starts a new transactions / session for e.g. a new job, and all the queries following should relate to this one? But how can a job itself then access the current session? E.g.
 -> "create new user" .. agenda starts transaction . the create new user creates some db entries, .. third one fails.. everythign shuld get rolled back? 
* What is if mongodb of agenda and the job's things are happening on different databases?
* What if the job does other stuff too, like sending mails or whatever..this is not part of this transaction then.
* Then there is this 60 second limit, so this limits this use case also, but still struggeling here with the first point anyway.

For the other operations, what would the expected benefit be? 

Don't get me wrong, I really want to support your submission, but I don't fully understand it yet?

davidpetrea:
Hello, any update on this?

PR #1389, opened 2021-10-12
fix: _unlockJobs should return when it resolves

esclapes:
Calling `resolve` does not stop the function execution. This can lead to
the [updateMany](https://github.com/agenda/agenda/blob/master/lib/agenda/stop.ts#L30) call happening even though the promise has resolved
and the promise is no longer awaiting.

For instance in this code:

```
await agenda.stop()
await agenda.close()
```

We may find that we already closed the connection when the stop code is
still trying to reach the Mongo DB

### DIFF
--- a/lib/agenda/stop.ts
+++ b/lib/agenda/stop.ts
@@ -23,7 +23,7 @@ export const stop = async function (this: Agenda): Promise<void> {
 
       if (jobIds.length === 0) {
         debug("no jobs to unlock");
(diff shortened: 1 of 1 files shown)


pdfowler:
The more you know™ ... thanks for the reference. I think the `else` approach would be overkill and this can be merged as-is. 

Now to see if this will help stabilize some intermittent issues in my test suite 🤔

PR #1401, opened 2021-11-07
feat: add generics while defining a job callback

### LINKED ISSUE #1400: Generic  Job not available in the callbakc of the define method.
When passing a callback for a `define` function, the `job` parameter doesn't have a Generic type in it. So, it becomes cumbersome when developing in typescript.

```
public static async callbackMethod(job: Job<GenericType>): Promise<void> {
  return Promise.resolve(job);
}
```

The above is not allowed.

I am creating a PR for the same.

SayakMukhopadhyay:
Closes #1400

### DIFF
--- a/lib/agenda/define.ts
+++ b/lib/agenda/define.ts
@@ -41,9 +41,9 @@ export interface DefineOptions {
   shouldSaveResult?: boolean;
 }
 
-export type Processor =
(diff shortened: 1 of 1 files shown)


svitekpavel:
I think the type of `data` in `JobAttributes` should also be changed.

Now:
`data?: T;`

Proposed change:
`data: T;`

Issue #893, opened 2019-12-15
Job stops working, continuously changing lockedAt every 10 minutes


antonCPU:
A job is configured to run every 30 minutes. But from time to time I see that the job stops working, apparently after some crash. But it should have been locked for some default time and then continue as per the schedule. But it only increases lockedAt time without actually running anything inside the job.

![image](https://user-images.githubusercontent.com/608935/70863875-755e0a80-1f55-11ea-9296-924cf5d5c860.png)

Please help to sort it out.

antonCPU:
Update: Could reliably reproduce the issue, looks like it's somehow connected to 2 jobs running and one of them fails. After around 5 cycles, everything halts and only lockedAt time is incrementing.

`const agenda = new Agenda().mongo(db.connection, 'jobs');

		const agendaJobs = [
			{
				jobName: 'test1',
				handler: (job, done) => {
					console.log('job1');
					//done();
				},
				scheduler: 'every 30 seconds',
			},
			{
				jobName: 'test2',
				handler: (job, done) => {
					console.log('job2');
					done();
				},
				scheduler: 'every 10 seconds',
			},
		];

		agenda.defaultLockLifetime(10000);

		agendaJobs.forEach(job => {
			agenda.define(job.jobName, job.handler);
		});

		agenda.on('ready', () => {
			agendaJobs.forEach(job => {
				agenda.every(job.scheduler, job.jobName);
			});
			agenda.start();
		});

		agenda.on('error', (err) => {
			console.log('Agenda error', err);
		});`

koresar:
I would be great if you could suggest a fix. Preferably as a pull request. Ping me, I'd be happy to review and merge it.
What do you think?

ikivela:
I had this problem too. Reverted Agenda back to version 2.0.1 to get my jobs running.

koresar:
@ikivela do you have an opportunity to fix this perhaps?

ikivela:
I am not so familiar with agenda that I could resolve this.

sonhle:
we had this with 2.3.0 too. @ikivela is it 2.1.0 working for you without this issue? Any workaround for this? Thanks.

ikivela:
Yes, 2.1.0 is working fine.

ikivela:
Workaround: Update nodejs to 12.16.1

I was using old version (10.16.2) which might have the effect.

ikivela:
[...]

PR #1420, opened 2022-01-11
Add job timeout

hrh47:
Related to #893

The issue can happen if a job is running and never ending. When running jobs number is greater than job concurrency number, a job will pop from queue and insert at the queue first position, which will be an endless loop.

### DIFF
--- a/lib/agenda/default-fail-on-timeout.ts
+++ b/lib/agenda/default-fail-on-timeout.ts
@@ -1,0 +1,20 @@
+import { Agenda } from ".";
+import createDebugger from "debug";
+
+const debug = createDebugger("agenda:defaultFailOnTimeout");
+
+/**
+ * Determine if the job should fail when it timeouts.
+ * Default is false
+ * @name Agenda#defaultFailOnTimeout
+ * @function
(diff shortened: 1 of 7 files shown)


harisvsulaiman:
@hrh47 thank you for your contribution

harisvsulaiman (approved):

harisvsulaiman:
@hrh47 please add a documentation to the readme file

simllll:
It's intended that thi sis a new parameter? Why don't just use the already exisitng lockTimeout?
In case you have two agenda workers running, it would do the same anyway (restarting a non locked job).

harisvsulaiman:
@simllll can you elaborate on the lockTimeout param

simllll:
@harisvsulaiman check out the implementation over here for this scenario:

https://github.com/hokify/agenda/blob/7bcdae1cfc4a395f47482f02168d1fdc85650e7e/src/JobProcessor.ts#L502

Here next to processing the job, a promise method that continually checks if the job is still alive is called. both are "racing" to complete. See the line with Promise.race()

Either the job finishes, or the check if the job lock has expires throws that it took too long.

Feel free to bring this change over!

Thanks for your work!

PR #1441, opened 2022-03-30
Feature: configuration for custom index on jobs collection

ysa23:
Added the ability to customize the index created by Agenda.

During our work with Agenda we found that the index agenda creates does not fit our use-case. We have about 500K scheduled jobs and the current index causes our database to choke. As an optimization we found that using a different index yields better performance - less load on the CPU & less scanned keys.

In general, it seems like a good idea to enable Agenda users to customize the index in case it is needed for a specific use-case.

Some of the results of the different index in our use case (MongoDB cloud, M40, 500K scheduled jobs):
CPU:
image
Profiler (keys examined):
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,6 +344,23 @@ Takes a `query` which specifies the sort query to be used for finding and lockin
 
 By default it is `{ nextRunAt: 1, priority: -1 }`, which obeys a first in first out approach, with respect to priority.
 
+### index(object)
+
+Optional. Takes an object which specifies the index to create on the jobs collection.
+
+The default index is:
+```json
+{
+  name: 1,
+  nextRunAt: 1,
+  priority: -1,
+  lockedAt: 1,
+  disabled: 1
+}
+```
+
+It is recommended to use the default index. This optional configuration is meant to enable specific optimizations that might be required under specific use-cases.
+
 ## Agenda Events
 
 An instance of an agenda will emit the following events:
--- a/lib/agenda/index.ts
+++ b/lib/agenda/index.ts
@@ -51,6 +51,7 @@ export interface AgendaConfig {
     collection?: string;
     options?: MongoClientOptions;
   };
+  index?: any,
 }
 
 /**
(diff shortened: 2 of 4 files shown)


harisvsulaiman:
Can you elaborate what indices were used and how it solved your problems.

ysa23:
@harisvsulaiman 
The index we are currently using is:
```
{
  name: 1,
  disabled: 1,
  lockedAt: 1,
  nextRunAt: 1,
}
```
From what we are seeing in the mongo metrics and profiler, this index reduces the number of keys being scanned when trying to acquire a lock on a job from thousands to a few dozens at worst.

simllll:
Can it be that you have one job definitino with thousand of jobs?
I had similar experience with that, and therefore have added this on our setup: https://github.com/hokify/agenda#performance

ysa23:
@simllll 
Thanks.
This is in fact our use-case - we have many instances of the same job.

We did try to add the index mentioned, and Mongo decided to pick (wrongfully) the default index mentioned instead of this one.
When we tested it (both using explain and in an actual environment) we see a major difference in the number of scanned keys between the two indexes - in favor of the one I mentioned.
The solution was to delete the default index and enforce Mongo to use the new one, which improved performance drastically (as can be seen in the graphs I attached).

simllll:
Ah okay, interesting. Seems like an issue with the query optimizer for me though, anyway to get this going, what do you think about another approach here to make things "easier": allow disabling the auto index creation.

This will give you all the freedom you need, but doesn't complicate the default setup for agenda.

ysa23:
@simllll 
Disable the auto creation is actually the first thing I tried to do (And does make more sense), but for some reason, when an index is not being created the tests fails and agenda does not start.
Looking at `db-init` I tried to simply not create the index according to a switch in the definitions, emit the ready event, but agenda does not seem to start.
I would be happy to do so - if anyone can provide guidance on how to actually make it work. I'll create a fork with this attempt so we can have something to discuss on.

simllll:
Thanks @ysa23, unfortuantely my time is quite limited, can't give you too much guidance right now, but what I can tell you is that [@hokify/agenda](https://www.npmjs.com/package/@hokify/agenda) is supporting this (look out for ensureIndex flag): source is over here https://github.com/hokify/agenda 
Maybe you can backport it from there?

ysa23:
@simllll 
I created a PR with the ability to disable the auto index creation:
https://github.com/agenda/agenda/pull/1444
If that looks good, I'll close this one

PR #1444, opened 2022-04-06
Feature: configuration for disabling auto index creation

ysa23:
Added the ability to prevent Agenda from creating its default index.

During our work with Agenda we found that the index agenda creates does not fit our use-case. We have about 500K scheduled jobs and the current index causes our database to choke. As an optimization we found that using a different index yields better performance - less load on the CPU & less scanned keys.

In general, it seems like a good idea to enable Agenda users to create their own index if they need it.

Per the discussion in a previous PR with @simllll , it was recommended to add the ability to disable the auto index creation instead of customizing the index:
https://github.com/agenda/agenda/pull/1441

Some of the results of the different index in our use case (MongoDB cloud, M40, 500K scheduled jobs):
CPU:

Profiler (keys examined):

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenda",
-  "version": "4.3.1",
+  "version": "4.2.1",
   "description": "Light weight job scheduler for Node.js",
   "main": "dist/cjs.js",
   "types": "dist/index.d.ts",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "agenda",
-  "version": "4.2.1",
+  "version": "4.3.1",
   "description": "Light weight job scheduler for Node.js",
   "main": "dist/cjs.js",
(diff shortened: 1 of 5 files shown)


harisvsulaiman (approved):
LGTM

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "agenda",
-  "version": "4.2.1",
+  "version": "4.3.1",
```
simllll, line 3: please don't do a version push in the PR itself, this will be done on release itself

simllll (commented):
except the package.json thing, LGTM

package.json
ysa23: Done

harisvsulaiman (approved):

simllll (approved):
Lgtm thanks for your contribution

harisvsulaiman:
@ysa23 LGTM . Merging this for now.

PR #1469, opened 2022-07-27
Fixed final occurrence not being calculated

Mainstay-Noah-Huppert:
Fixes a bug where the final occurrence of a job scheduled with a cron string would fail to be calculated. 

This would happen because there existed a check to ensure that the next calculated occurrence was always before (inclusive) the end date of the job. However the end date value this check compared to was the end date, but at time 00:00:00. Instead of at the time stored for the job.

For example imagine there is a job with the cron string `0 20 * * *`. This specifies the job should run daily at 20:00:00 (8 PM). If you want this job to run up until (inclusive) 2022-07-31 you might set an `endDate` for the job to be `2022-07-31 20:00:00.000Z`. During job invocation before the final run of the job, aka at `2022-07-30 20:0000.000Z`, Agenda goes to calculate the next occurrence. It would find the next occurrence date to be `2022-07-31 20:00:00.000Z`. It would then compare this to the stored end date, but with a time of 00:00:00. Resulting in the comparison `2022-07-31 20:00:00.000Z < 2022-07-31 00:00:00.000Z`. This comparison would fail and in your Agenda job you would find:

```json
"failCount" : 1,
"failReason" : "failed to calculate nextRunAt due to invalid repeat interval",
```

The end date parameter is inclusive according to the docs on `endDate`:

> options.endDate: Date the job should not repeat after the endDate. The job can run on the end-date itself, but not after that.

Therefore Agenda's logic should find this final occurrence to be valid. However due to the logical bug of stripping the time of end date before the comparison Agenda found the final occurrence to be invalid.

### DIFF
--- a/lib/job/compute-next-run-at.ts
+++ b/lib/job/compute-next-run-at.ts
@@ -83,13 +83,8 @@ export const computeNextRunAt = function (this: Job): Job {
       }
 
       // If endDate is less than the nextDate, set nextDate to null to stop the job from running further
-      if (endDate) {
-        const endDateDate: Date = moment
-          .tz(moment(endDate).format("YYYY-MM-DD"), timezone!)
-          .toDate();
-        if (nextDate > endDateDate) {
-          nextDate = null;
-        }
+      if (endDate && nextDate > endDate) {
+        nextDate = null;
       }
 
       this.attrs.nextRunAt = nextDate;


manarhusrieh:
@koresar May you review this PR

NurdinDev:
I'm facing the same issue, the issue just appears with cron expression and with endDate or startDate
after digging deep into this issue, I found the following:

- First I'm sending this information:
```
{
       "expression": "*/1 * * * *",
        "options": {
                  "endDate": "2022-10-20T13:18:20.570Z"
            }
}


"failReason" : "failed to calculate nextRunAt due to invalid repeat interval"
```

- The `cronTime.next().toDate()` will return the date with minutes, like this: `2022-10-20T13:17:00.000Z ` as I noticed the [cron-parser](https://github.com/harrisiirak/cron-parser) `next().toDate()` is returning dateTime with **minutes**, but after formatting endDate here `moment(endDate).format("YYYY-MM-DD")`  the result is `2022-10-20T00:00:00.000Z` which is a date without **minutes**, therefore what will happen now when the comparison happens between the two dates here: https://github.com/agenda/agenda/blob/b0850345b3a5a61a3b5fb888293a3c3a20c9fc9e/lib/job/compute-next-run-at.ts#L90
the result will always be truthy, because `2022-10-20T13:17:00.000Z > 2022-10-20T00:00:00.000Z`

I'm guessing the solution to this bug by formatting the date similar to the format of cron-parser
something like `moment(endDate).format("YYYY-MM-DD HH:mm")`

harisvsulaiman:
closed in favour of https://github.com/agenda/agenda/pull/1497/

PR #1530, opened 2023-04-27
Remove saveResults examples from README.md

luizcarlos1405:
The current docs do not include a saveResults or setSaveResults and they do not work on V6.x

image


luizcarlos1405:
Oh, it seems this repo doesn't have version 6?

Issue #1502, opened 2022-11-08
bring agendaTS into agenda main repo


simllll:
As discussed and suggested here https://github.com/agenda/agenda/issues/1082 the idea is to bring [agendaTS](https://github.com/hokify/agenda) as new major version of agenda main.

Checklist (WIP):
- [x] try to create a PR that is based on the current agenda master, and try to keep as many commits from the agendaTS to have a change history. in case of merge conflicts, keep the agendaTS code.
- [x] rename everything back to agenda
- [ ] improve docs of new features
- [ ] write down list of breaking changes and transition guide

New features in agendaTS:
- touch() can have an optional progress parameter (0-100)
- job forking in seperate threads (https://github.com/hokify/agenda#sandboxed-worker---use-child-processes)
- getRunningStats()

List of currently known breaking changes:
- not supported in agendaTS: Feat: Add top level disable and enable (https://github.com/agenda/agenda/pull/1109) thanks @pdfowler
- not supported in agendaTS: shouldSaveResult (see  https://github.com/hokify/agenda/issues/29)
- not supported in agendaTS: Feat: Add drain() method for graceful process shutdowns (https://github.com/agenda/agenda/pull/1488) thanks @nmehmedrnd
- agenda default export (cjs?) - check compatiblity
- not supported in agendaTS: no support for "non public api" export of internal db connection (there is currently no _collection workaround: `const collection = agenda._collection?.collection || genda._collection || agenda.db?.collection;` see https://github.com/hokify/agenda/issues/27
- not supported in agendaTS: Add ability to schedule jobs with startDate, endDate, and skipping (https://github.com/agenda/agenda/pull/361) thanks @suryanaik
- define() config paramter moved from 2nd position to 3rd

feel free to add comments, ideas and other things you come up with :)

harisvsulaiman:
What about tests?

harisvsulaiman:
@simllll I hope this will be a starting point.

harisvsulaiman:
@simllll Is there any update on this issue?

simllll:
Sorry for still not working on it, I have it still in my list, but couldn't really find time yet. Hopefully within the next few weeks though!

harisvsulaiman:
@simllll Is this on the agenda ? (Pun intended)

simllll:
It's still on my list, but unfortunately didn't find time yet... Maybe I can start working on it next weekend, to get things rolling... 🤞

harisvsulaiman:
[...]

PR #1536, opened 2023-05-24
feat: add top-level `agenda.disable` and `agenda.enable` methods

pdfowler:
I noticed my work was listed as a blocker on #1502, so here is my original implementation of the `enable` and `disable` methods ported to the agendaTS codebase.

* this code has been migrated over from agenda/agenda#1109
* migrates old methods from individual files to the top level Agenda Class
* test suite updated with original test suite content from original fork

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 import { EventEmitter } from 'events';
 import * as debug from 'debug';
 
-import type { Db, Filter, MongoClientOptions, Sort } from 'mongodb';
+import type { Db, Filter, MongoClientOptions, Sort, UpdateResult } from 'mongodb';
 import { SortDirection } from 'mongodb';
 import { ForkOptions } from 'child_process';
 import type { IJobDefinition } from './types/JobDefinition';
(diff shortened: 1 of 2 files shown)


simllll:
Oh great, thanks for your work.

One thing to mention: I've introduced the JobDbRepository to "abstract" away the datbaes layer,
in your implementation the database querys (.collection.updateMany()) are now implemented directly in the index.ts file, can you refactor it so that you just move the "updateMany" things into the JobDbRepository and call the method instead.
What's the idea behind: have adapter/database logic abstracted away from application/business logic. this helps to easier read and test code, but also gives the possibility to switch the database layer, without changing any application code/logic.

Thanks again for your contribution.

harisvsulaiman:
@simllll I love the idea that we could switch out db layer in the future. 
But I think we can support many features like (watching for new jobs and even job defs) if we stick to mongo

simllll:
> @simllll I love the idea that we could switch out db layer in the future. But I think we can support many features like (watching for new jobs and even job defs) if we stick to mongo

Yep, my main concern is to keep "business" logic and adapter logic seperat, this helps maintating a cleaner code :) But it would be a possible "side effect" that allows swichting the layer too, but this is for sure another topic we can discuss in the future ;-)

pdfowler:
Took a quick pass at abstracting things out. I like the idea of making it more DB agnostic, but I imagine the transition will be challenging, especially for something like this which is so interconnected with mongo syntax.

I'll try to take another pass at cleaning up the formatting in the next few days

PR #1597, opened 2025-03-21
Feature/update mongo drivers

marco-bertelli:
Hi team i take an old and already approved PR to upgrade to new mongoDrivers to support MongoDb 8.

can you approve that PR so we dosen't have to migrate away from agenda?

Thanks in advice

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,20 +1,20 @@
 {
   "name": "agenda",
-  "version": "4.4.0",
+  "version": "5.0.1",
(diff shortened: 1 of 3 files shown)


marco-bertelli:
hi team i see that you approved the original pr but maybe the author is not active if you want is the same thing :)

@adrian-BOS

busterbogheart:
Is this still in progress?  With the coming Mongo deprecations, this will be necessary for anyone going to v8.

PR #1670, opened 2026-02-11
Fix: add "default" export condition to all packages

ysa23:
The exports map in all packages only specifies the "import" condition. This prevents CommonJS projects from using require() to load these packages, since Node.js require() matches "require" or "default" conditions — not "import".

With require(esm) now stable in Node.js 20.19+, 22.12+, and 24+, adding a "default" condition allows CJS projects to require() these ESM packages without needing workarounds (patches, dynamic import, or module name mappers).

The "default" condition points to the same dist files as "import", so there is no change in behavior for ESM consumers.

## Description

Brief description of changes.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Updated documentation if needed

### DIFF
--- a/.changeset/add-default-export-condition.md
+++ b/.changeset/add-default-export-condition.md
@@ -1,0 +1,12 @@
+---
+"agenda": patch
+"@agendajs/mongo-backend": patch
+"@agendajs/postgres-backend": patch
+"@agendajs/redis-backend": patch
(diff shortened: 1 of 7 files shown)


simllll (approved):
Thanks, sounds good to me

Issue #1698, opened 2026-04-27
Does Agenda provide a way to show the StackTrace in the fail reason


felixwoestmann:
Hello, 

I was wondering whether Agenda supports more insightful logs, namely showing more information, such as the StackTrace in the reason for a failure. This would make debugging easier.

<img width="591" height="572" alt="Image" src="https://github.com/user-attachments/assets/da3599be-dac9-4587-9a3e-4e054ba3feae" />

Nian-PM:
I opened #1700 to preserve stack traces in `job.attrs.failReason` when a job fails with an `Error`.

The change:
*   Keeps string fail reasons unchanged.
*   Stores `error.stack` when available, falling back to `error.message`.
*   Includes unit tests and a patch changeset.

simllll:
Closed with #1705

PR #1700, opened 2026-05-03
fix: preserve error stacks in fail reason

Nian-PM:
## Summary

Preserves stack traces when a job fails with an `Error`.

Previously, `job.fail(error)` stored only `error.message` in `job.attrs.failReason`, which made Agendash and stored job state less useful for debugging. This change stores `error.stack` when available and falls back to `error.message` when a stack is unavailable.

String fail reasons are unchanged.

## Related

Helps address #1698.

## Changeset

Included a patch changeset for the `agenda` package.

## Verification

- `pnpm --filter agenda test`
- `pnpm --filter agenda build`
- `pnpm --filter agenda typecheck`
- `pnpm lint`
- `pnpm build`

### DIFF
--- a/.changeset/preserve-error-stacks.md
+++ b/.changeset/preserve-error-stacks.md
@@ -1,0 +1,5 @@
+---
+"agenda": patch
+---
+
+Preserve Error stack traces in `job.attrs.failReason` when jobs fail with an Error.
--- a/README.md
+++ b/README.md
@@ -1580,7 +1580,8 @@ await job.save();
 Sets `job.attrs.failedAt` to `now`, and sets `job.attrs.failReason` to `reason`.
 
 Optionally, `reason` can be an error, in which case `job.attrs.failReason` will
-be set to `error.message`
+be set to `error.stack`. If the error has no stack, Agenda falls back to
+`error.message`.
 
 ```js
 job.fail('insufficient disk space');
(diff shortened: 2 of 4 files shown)


simllll:
I see your point, but I'm hesitant for a couple of reasons:

1. Breaking change — this modifies an existing property's shape, so anyone currently reading failReason as a string would break. 
2. Payload size — stack traces can get quite large (and recursive in some error scenarios), and this gets persisted to the database on every failure. That's a cost every user would pay.                                                                             

For Agenda itself, I'd prefer to keep task outcomes out of scope. If you need richer error context, there are a few approaches that work well today:                                                                                                                

  - Attach an error / fail listener — you get the full Error object and can persist the stack wherever makes sense (a separate table,
   file, etc.)
  - Pipe errors into Sentry, Datadog, or whatever observability tool you already use                                                 
  - Wrap your job handler and handle structured error capture there                                                                  

  That keeps the core lean and gives you full control over how much detail you store and where.

Nian-PM:
@simllll 

Thanks for the detailed explanation. That makes sense to me.

I agree that keeping `failReason` small and preserving the existing behavior is the better default for Agenda core, especially given the persistence cost and compatibility concerns.

I opened #NEW_PR_NUMBER as a follow-up docs PR to show the recommended `fail` listener pattern for capturing richer error context outside Agenda. I'll close this implementation PR.

PR #1703, opened 2026-05-05
fix mongo data filtering for ObjectId values

### LINKED ISSUE #1702: Mongo backend cannot cancel/query jobs by ObjectId field inside data
## Description
When using the MongoDB backend, Agenda cannot cancel a job by matching an `ObjectId` value inside the job `data`.

A clear description of the bug.

For example, if a job is created with data like:

```ts
{
  id: new ObjectId("64a000000000000000000001")
}
```

then calling `agenda.cancel()` with the same `ObjectId` inside `data` does not remove the job.

I think this happens because `flattenDataFilter()` recursively flattens any non-null object except `Date`. Since MongoDB `ObjectId` is also an object in JavaScript, it is treated as a nested object instead of being preserved as a BSON scalar value.

## Steps to Reproduce

1. Create a job with an `ObjectId` inside the job data:

```ts
import { ObjectId } from "mongodb";

const id = new ObjectId("64a000000000000000000001");

await agenda.schedule("in 1 hour", "test-job", {
  id
});
```

2. Try to cancel the job by matching the same data field:

```ts
const removed = await agenda.cancel({
  name: "test-job",
  data: {
    id
  }
});

console.log(removed);
```

3. Check the return value and the database collection.

## Expected Behavior

The job should be removed.

```ts
removed === 1
```

The generated MongoDB query should effectively match:

```ts
{
  name: "test-job",
  "data.id": ObjectId("64a000000000000000000001")
}
```

## Actual Behavior

The job is not removed.

```ts
removed === 0
```

The job remains in the Agenda MongoDB collection.

It looks like the `ObjectId` value is being flattened as if it were a nested plain object, instead of being preserved as a BSON `ObjectId`.

## Environment

- Node.js version: v20., running in Docker image `node:20-alpine`
- Agenda version: `agenda@6.2.4`
- Backend: MongoDB
- Agenda Mongo backend version: `@agendajs/mongo-backend@4.0.1`
- MongoDB driver version: `mongodb@7.1.1`
- OS: Linux

## Additional Context

This may also affect other methods that use the same `data` filtering logic, such as:

- `queryJobs()`
- `removeJobs()`
- `disableJobs()`
- `enableJobs()`

A possible fix would be to recursively flatten only plain objects, while treating BSON values like `ObjectId` as terminal values.

For example:

```ts
function isPlainObject(value: unknown): value is Record<string, unknown> {
  if (value === null || typeof value !== "object") {
    return false;
  }

  const prototype = Object.getPrototypeOf(value);
  return prototype === Object.prototype || prototype === null;
}
```

Then `flattenDataFilter()` could recurse only when `isPlainObject(value)` is true.

NandPatel1292:
## Summary

Fixes Mongo backend filtering when job `data` contains BSON values such as `ObjectId`.

Previously, `flattenDataFilter()` recursively flattened any non-null object except `Date`. Since MongoDB `ObjectId` is also an object in JavaScript, filters like this did not match correctly:

```ts
{
  data: {
    userId: new ObjectId(...)
  }
}
```

Instead of preserving the value as:

```ts
{
  "data.userId": ObjectId(...)
}
```

the filter could incorrectly treat the `ObjectId` as a nested object.

## Changes

- Add an `isPlainObject()` helper.
- Only recursively flatten plain objects.
- Preserve `ObjectId` and other non-plain object values as terminal filter values.
- Add a regression test covering `queryJobs()` and `removeJobs()` with `ObjectId` inside job data.

## Related issue

Fixes #1702

## Test plan

```bash
pnpm --filter @agendajs/mongo-backend exec vitest run test/mongo-backend.test.ts -t "ObjectId values when filtering job data"
pnpm --filter @agendajs/mongo-backend exec vitest run test/mongo-backend.test.ts -t "MongoDB-specific features"
pnpm --filter @agendajs/mongo-backend typecheck
pnpm --filter @agendajs/mongo-backend build
```
The focused MongoDB-specific tests for this change pass.

### DIFF
--- a/.changeset/clean-bulldogs-occur.md
+++ b/.changeset/clean-bulldogs-occur.md
@@ -1,0 +1,5 @@
+---
+"@agendajs/mongo-backend": patch
+---
+
+fix mongo data filtering for ObjectId values
--- a/packages/mongo-backend/src/MongoJobRepository.ts
+++ b/packages/mongo-backend/src/MongoJobRepository.ts
@@ -39,6 +39,18 @@ function escapeRegex(str: string): string {
 }
 
 /**
+ * Check if a value is a plain object (not an array, date, or other non-plain type)
+ */
+function isPlainObject(value: unknown): value is Record<string, unknown> {
+	if (value === null || typeof value !== 'object') {
+		return false;
+	}
+
+	const prototype = Object.getPrototypeOf(value);
+	return prototype === Object.prototype || prototype === null;
+}
+
+/**
  * Flatten a data filter object into MongoDB dot-notation keys.
  * This enables partial matching on the data subdocument.
  *
@@ -51,8 +63,8 @@ function flattenDataFilter(data: Record<string, unknown>, prefix = 'data'): Reco
 	const result: Record<string, unknown> = {};
 	for (const [key, value] of Object.entries(data)) {
 		const fullKey = `${prefix}.${key}`;
-		if (value !== null && typeof value === 'object' && !Array.isArray(value) && !(value instanceof Date)) {
-			Object.assign(result, flattenDataFilter(value as Record<string, unknown>, fullKey));
+		if (isPlainObject(value)) {
+			Object.assign(result, flattenDataFilter(value, fullKey));
 		} else {
 			result[fullKey] = value;
 		}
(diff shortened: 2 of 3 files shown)


simllll:
do we have tests for Array.isArray(value) and value instanceof Date too?

NandPatel1292:
Thanks for pointing this out. I added regression coverage for both array and Date values in data filters to make sure they continue to be treated as terminal values and are not recursively flattened.

simllll (approved):
thanks (y)

PR #1718, opened 2026-06-13
Fix queue order, @Every scheduling, retry reset, and a Postgres SQL injection; harden Mongo unique()

### LINKED ISSUE #1719: Jobs queued in the same tick run in reverse priority order
## Description

`JobProcessingQueue` is read from the end of the array — `returnNextConcurrencyFreeJob` iterates it in reverse. But `insert()` puts a job that should sort before everything else (higher priority, or earlier `nextRunAt`) at the front of the array with `unshift`, which is the slot that gets read last. So when several jobs are enqueued in the same processing tick, they come out in the wrong order.

## Steps to Reproduce

1. Define two jobs and enqueue them in the same tick, one with higher priority than the other.
2. Observe the order in which they're pulled for execution.

## Expected Behavior

The higher-priority (or earlier `nextRunAt`) job runs first.

## Actual Behavior

It runs last — the order is reversed.

## Fix

`packages/agenda/src/JobProcessingQueue.ts`: push to the end of the array instead of `unshift`. Opened in #1718 with a regression test.

## Environment

- Package: `agenda` (core)
- Branch: `main`

Noethix55555:
Found a few bugs in the TypeScript packages and fixed them. Each one has a test, and there's a longer write-up in BUGS.md.

The processing queue ran jobs in the wrong order. It's read from the end of the array, but `insert()` put the highest-priority (or earliest) job at the front, so jobs queued in the same tick came out backwards. Now it pushes to the end.

`@Every` jobs registered after start never got scheduled. The scheduler waited on a one-shot `ready` event, which has already fired by the time you do `await agenda.ready` and then call `registerJobs`. Switched it to the `ready` promise so it works whether or not the agenda has started yet.

Retries stopped working on recurring jobs. `failCount` drives the backoff attempt number but was never reset after a success, so once a job burned through its retries it never backed off again. It's reset on success now.

The Postgres backend had a SQL injection in `unique()`. The JSON path and column name were dropped straight into the query, so a crafted key could rewrite the `WHERE` clause behind the upsert and, since the `UPDATE` has no `LIMIT`, overwrite other jobs. The path is bound as a parameter now and the column name is checked against the real table columns.

The fork worker in the example helper imports the job file by absolute path, which breaks on Windows with `ERR_UNSUPPORTED_ESM_URL_SCHEME`. Wrapped it in `pathToFileURL`. That's not shipped code, just the bit people copy.

Last one isn't really a bug, more a suggestion. Mongo's `unique()` takes a raw query filter and passes it through as-is, which is fine until the key gets built from user input. `$where`, `$function`, `$expr` and `$accumulator` have no business in a uniqueness key, so it rejects them now. Normal equality and comparison filters still work.

Changesets are in for agenda, postgres-backend and mongo-backend. All tests pass and lint is clean.

Closes #1719, #1720, #1721, #1722, #1723. Addresses #1724.

### DIFF
--- a/.changeset/fix-scheduler-queue-retry.md
+++ b/.changeset/fix-scheduler-queue-retry.md
@@ -1,0 +1,9 @@
+---
+'agenda': patch
+---
+
+Fix three scheduling bugs:
+
+- The processing queue placed the highest-priority (or earliest) job in the slot that runs last, so jobs queued in the same tick ran in reverse priority order.
+- `@Every` jobs registered after the agenda was already ready were never scheduled, because the scheduler relied on a one-shot `ready` event that had already fired.
+- A recurring job's backoff retry counter was never reset after a successful run, so a job that had exhausted its retries earlier in its lifetime stopped retrying.
--- a/.changeset/mongo-unique-operator-guard.md
+++ b/.changeset/mongo-unique-operator-guard.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Reject server-side-evaluation operators (`$where`, `$function`, `$expr`, `$accumulator`) in `unique()` constraints. They have no use in a uniqueness key and could allow query injection when the key is built from untrusted input. Ordinary equality and comparison constraints are unchanged.
--- a/.changeset/postgres-unique-sql-injection.md
+++ b/.changeset/postgres-unique-sql-injection.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/postgres-backend': patch
+---
+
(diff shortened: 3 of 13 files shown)


cclauss (changes requested):
GitHub account was created today.

Noethix55555:
it was not???

koresar:
Please submit 5 bug fixes in 5 different PRs. This type of PRs is not sustainable.

Noethix55555:
Closing in favor of the per-issue split PRs, one bug each: #1725 (queue order), #1726 (@Every after ready), #1727 (failCount reset), #1728 (Postgres unique() SQL injection), #1729 (fork worker file URL), #1730 (Mongo unique() operators). This combined PR is now fully superseded by those.

PR #1725, opened 2026-06-14
fix(agenda): correct processing queue insertion order

### LINKED ISSUE #1719: Jobs queued in the same tick run in reverse priority order

Noethix55555:
Fixes #1719.

`JobProcessingQueue.insert` placed the highest-priority (or earliest `nextRunAt`) job at the front of the array with `unshift`. The queue is read from the end (`returnNextConcurrencyFreeJob` scans in reverse), so that job ran last. Jobs queued in the same tick came out in reverse priority order.

The fix pushes the no-earlier-match job to the end instead.

Test inserts jobs in both orders and asserts the higher-priority and earlier `nextRunAt` job runs first. It fails on the previous code and passes with the fix.

### DIFF
--- a/.changeset/fix-queue-order.md
+++ b/.changeset/fix-queue-order.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Fix processing queue order. `JobProcessingQueue.insert` placed the highest-priority (or earliest `nextRunAt`) job in the slot that runs last, because the queue is read from the end. Jobs queued in the same tick now run in the correct priority order.
--- a/packages/agenda/src/JobProcessingQueue.ts
+++ b/packages/agenda/src/JobProcessingQueue.ts
@@ -79,8 +79,9 @@ export class JobProcessingQueue {
 		});
 
 		if (matchIndex === -1) {
-			// put on left side of the queue
-			this._queue.unshift(job);
+			// no existing job should run before this one: it is the earliest /
+			// highest-priority entry, so it belongs at the right (processed first)
+			this._queue.push(job);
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified against the queue invariant: left-to-right decreasing nextRunAt with the reader scanning from the end. `matchIndex === -1` means the inserted job beats every queued job (earliest/highest-priority), and `unshift` placed exactly that job in the last-processed slot — the asymmetric test (low,high fails / high,low passes on old code) proves it. `push` restores the invariant; mixed-case and empty-queue behavior unchanged. Rebased onto current main (incl. today's #1743/#1750/#1759 batch), agenda 178 + mongo 259 green, lint clean.

PR #1726, opened 2026-06-14
schedule @Every jobs registered after agenda is ready

### LINKED ISSUE #1720: @Every jobs registered after start are silently never scheduled
## Description

`@Every` jobs registered after the agenda has already started never get scheduled. The decorator scheduler waits on a one-shot `ready` event (`agenda.once('ready', ...)`). Agenda emits `ready` once, early in the constructor, so by the time you do `await agenda.ready` and then call `registerJobs`, the event has already fired and the listener never runs. Nothing throws — the `@Every` jobs just silently never get created.

The code's own comment says it should "schedule immediately if agenda is already started", which `once('ready')` can't do.

## Steps to Reproduce

1. `await agenda.ready` (or `await agenda.start()`).
2. Call `registerJobs` for a controller that has `@Every` jobs.
3. Check the jobs collection.

## Expected Behavior

The `@Every` jobs are scheduled.

## Actual Behavior

Nothing is scheduled and no error is raised.

## Fix

`packages/agenda/src/decorators/register.ts`: use the `ready` promise instead of the one-shot event, so it works whether or not the agenda has started yet. Opened in #1718 with a regression test.

## Environment

- Package: `agenda` (core), decorator/`registerJobs` API
- Branch: `main`

Noethix55555:
Fixes #1720.

`@Every` jobs were scheduled from a `once('ready')` listener. Agenda emits `ready` once, in the constructor. Code that does `await agenda.ready` before calling `registerJobs` attaches the listener after the event already fired, so the jobs are never scheduled and nothing throws. The code's own comment says it should schedule immediately if agenda is already started, which `once('ready')` cannot do.

The fix uses the `ready` promise, which stays resolved after the event.

Test awaits `agenda.ready`, registers a controller with an `@Every` job, and checks it was saved.

### DIFF
--- a/.changeset/fix-every-after-ready.md
+++ b/.changeset/fix-every-after-ready.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Fix `@Every` jobs registered after the agenda was already ready never being scheduled. The scheduler relied on a one-shot `ready` event that had already fired, so it now uses the `ready` promise, which stays resolved.
--- a/packages/agenda/src/decorators/register.ts
+++ b/packages/agenda/src/decorators/register.ts
@@ -149,13 +149,15 @@ function scheduleEveryJobsOnReady(agenda: Agenda, pendingJobs: PendingEveryJob[]
 		}
 	};
 
-	// If agenda is already started, schedule immediately
-	// Otherwise, wait for the 'ready' event
-	agenda.once('ready', () => {
-		scheduleJobs().catch(err => {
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified on main: 'ready' fires exactly once at backend connect (index.ts:330) and the `ready` promise resolves on it and stays resolved — so `once('ready')` after readiness silently drops every @Every job, exactly as described. The promise-based fix preserves both orderings and routes failures to the 'error' event. Rebased onto current main, agenda suite green (185), lint clean.

PR #1727, opened 2026-06-14
reset backoff failCount on success

### LINKED ISSUE #1721: Backoff failCount is never reset on success, so recurring jobs stop retrying
## Description

A recurring job's backoff retry counter is never reset after a successful run. `fail()` increments `failCount`, and `handleRetry()` uses `failCount` as the attempt number to compute the next backoff. A successful run never resets `failCount`, so once a recurring job has exhausted its retries earlier in its lifetime, every later failure is treated as already out of retries and gets no backoff — retries silently stop working for the rest of the job's life.

## Steps to Reproduce

1. Define a recurring job with retry/backoff configured.
2. Let it fail enough times to exhaust its retries.
3. Let it succeed once.
4. Make it fail again.

## Expected Behavior

After the success, the retry counter is back to zero and the next failure schedules a normal backoff retry.

## Actual Behavior

The old `failCount` is still there, so the next failure is treated as out of retries and no retry is scheduled.

## Fix

`packages/agenda/src/Job.ts`: reset `failCount` on success. (`0` persists cleanly and loads back as `undefined`, so jobs that never failed are unaffected.) Opened in #1718 with a regression test.

## Environment

- Package: `agenda` (core)
- Branch: `main`

Noethix55555:
Fixes #1721.

`fail()` increments `failCount`, and `handleRetry()` uses it as the attempt number. A successful run never reset it. A recurring job that exhausts its retries once and then succeeds keeps the old count, so its next failure is treated as already out of retries and gets no backoff.

The fix resets `failCount` on success. `0` persists cleanly and loads back as `undefined`, so a job that never failed is unaffected.

Test exhausts the retry budget, runs one success, fails again, and checks a new retry is scheduled.

### DIFF
--- a/.changeset/fix-backoff-failcount-reset.md
+++ b/.changeset/fix-backoff-failcount-reset.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Reset a job's backoff retry counter after a successful run. A recurring job that exhausted its retries earlier in its lifetime kept the old `failCount`, so a later failure was treated as already out of retries and stopped retrying.
--- a/packages/agenda/src/Job.ts
+++ b/packages/agenda/src/Job.ts
@@ -693,6 +693,14 @@ export class Job<DATA = unknown | void> {
 
 			this.attrs.lastFinishedAt = new Date();
 
+			// Reset the failure counter on success so a later failure starts a
+			// fresh retry/backoff sequence instead of inheriting the old count.
+			// This matters for recurring jobs that recover between runs.
+			// 0 round-trips back to `undefined` via computeJobObj on reload.
+			if (this.attrs.failCount) {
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified: fail() only ever increments (Job.ts:336) and handleRetry derives the attempt number from failCount (`attempt: failCount || 1`), so exhaust-then-recover permanently disabled future retries — real bug. Reset-on-success is the right semantic (attempt counter for the current sequence, not a lifetime stat). computeJobState doesn't read failCount, so no state side effects. Note: I corrected the stale code comment on your branch — since #1783, a persisted 0 loads back as 0 (not undefined), which is behaviorally identical given the falsy coercion at all consumers. Rebased onto current main, agenda 179 + mongo 259 green, lint clean.

PR #1728, opened 2026-06-14
prevent SQL injection via unique() constraint keys

### LINKED ISSUE #1722: SQL injection via unique() constraint keys (Postgres backend)
## Description

The PostgreSQL backend builds the `unique()` constraint lookup by concatenating the constraint keys straight into the SQL. The values are bound, but the JSON path and the column name are not:

```js
conditions.push(`data->>'${dataPath}' = $${paramIndex++}`);  // JSON path key
conditions.push(`${columnName} = $${paramIndex++}`);          // column name
```

Apps commonly build `unique()` keys from request data ("one job per user"), so the key can be attacker-controlled. A key like `data.id' OR '1'='1` rewrites the `WHERE` clause of the existence check that drives the upsert. The matching `UPDATE` has no `LIMIT`, so it can overwrite every job's `data` and `next_run_at` — and if a target job's handler uses `data` to run a command, build a path, or make a request, that turns into command execution, SSRF, or path traversal in the application.

## Steps to Reproduce

1. Save a job with a `unique()` constraint whose key is `"data.id' OR '1'='1"`.
2. Inspect the generated SQL / the rows it matches.

## Expected Behavior

The key is treated as data; the lookup only matches the intended job.

## Actual Behavior

The key changes the query's logic and the upsert can match and overwrite unrelated jobs.

## Fix

`packages/postgres-backend/src/PostgresJobRepository.ts`: bind the JSON path as a parameter, and validate the column name against the table's real columns (identifiers can't be bound). Opened in #1718 with a regression test that uses a mock pool, so it needs no live database.

## Environment

- Package: `@agendajs/postgres-backend`
- Branch: `main`

Noethix55555:
Fixes #1722.

The unique-constraint lookup bound the values but put the keys straight into the SQL:

```js
conditions.push(`data->>'${dataPath}' = $${paramIndex++}`);  // JSON path key
conditions.push(`${columnName} = $${paramIndex++}`);         // column name
```

A key like `data.id' OR '1'='1` rewrites the WHERE clause of the existence check that drives the upsert. The matching UPDATE has no LIMIT, so it can overwrite every job's `data` and `next_run_at`, and force them to run immediately.

The fix binds the JSON path as a parameter, and validates the column name against the table's columns since identifiers cannot be bound.

Test uses a recording mock pool: it checks the payload is bound rather than present in the SQL text, that unknown columns are rejected, and that legitimate constraints still work.

### DIFF
--- a/.changeset/postgres-unique-sql-injection.md
+++ b/.changeset/postgres-unique-sql-injection.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/postgres-backend': patch
+---
+
+Prevent SQL injection through `unique()` constraint keys. JSON paths are now bound as query parameters and column names are validated against the table's columns, instead of being concatenated into the SQL.
--- a/packages/postgres-backend/src/PostgresJobRepository.ts
+++ b/packages/postgres-backend/src/PostgresJobRepository.ts
@@ -54,6 +54,9 @@ const UNIQUE_QUERYABLE_COLUMNS = new Set<string>([
 	'fork',
 	'last_modified_by',
 	'debounce_started_at',
+	'start_date',
+	'end_date',
+	'skip_days',
 	'created_at',
 	'updated_at'
 ]);
--- a/packages/postgres-backend/src/PostgresJobRepository.ts
+++ b/packages/postgres-backend/src/PostgresJobRepository.ts
@@ -27,6 +27,38 @@ import {
 const log = debug('agenda:postgres:repository');
 
 /**
+ * Columns that may be referenced directly (i.e. not via a `data.` JSON path) in a
+ * `unique` constraint. SQL identifiers cannot be passed as bound parameters, so a
+ * caller-supplied key is validated against this allowlist before it is ever placed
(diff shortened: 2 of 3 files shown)


simllll:
Verified this against real infrastructure (the mock-pool test can't catch SQL-validity issues):

- ✅ `data->>$N` with an untyped bound parameter works on Postgres 16 via node-pg (no `operator is not unique` ambiguity)
- ✅ `pool` + `ensureSchema: false` are valid config, so the recording-mock test construction is sound
- ✅ The conditions array feeds both the existence SELECT and the following UPDATE's WHERE, so both statements are covered by the fix
- ✅ Interleaved `paramIndex++` ordering matches the `params.push(dataPath, value)` order

One gap: the allowlist predated #1789 (merged today), which added `start_date`, `end_date` and `skip_days` to the schema — I've pushed a commit to your branch adding them (maintainer edits were enabled, hope that's fine). Thanks for the report and the clean fix — the identifier-allowlist + bound-JSON-path approach is exactly right.

PR #1730, opened 2026-06-14
reject unsafe operators in unique() constraints

### LINKED ISSUE #1724: Reject server-side-eval operators in Mongo unique() constraints (improvement)
## Summary

This is a hardening suggestion, not a bug. The Mongo backend's `unique()` takes a raw query filter and passes it through to the query as-is, which is the documented behaviour. That's fine until an app builds the unique key from user input — then operators like `$where`, `$function`, `$expr`, and `$accumulator` can run server-side code, and `$ne` / `$regex` / `$gt` can make the lookup match a different job (the same shape as the Postgres `unique()` issue, just object-based instead of string-based).

The `data` query path is already safe because it flattens operators into literal keys, but `unique` isn't flattened.

## Suggestion

Reject the server-side-evaluation operators (`$where`, `$function`, `$expr`, `$accumulator`) inside a `unique()` constraint before building the query, since they have no use in a uniqueness key. Ordinary equality and comparison constraints stay untouched.

Implemented in #1718 with unit + integration tests, for discussion. Happy to drop it if you'd rather leave `unique()` fully pass-through and document the expectation instead.

Noethix55555:
Fixes #1724.

This is a hardening suggestion rather than a library bug. MongoDB queries are objects, and `unique()` is documented to take a query filter, so the backend passes it through as designed. But if an app builds `unique` keys from untrusted input, `$where`, `$function`, `$expr`, and `$accumulator` run server-side code. The data query path is already safe because it flattens operators into literal keys, but `unique` is not flattened.

Since the code-evaluation operators have no use in a uniqueness key, the backend now rejects them before building the query. Ordinary equality and comparison constraints (`$ne`, `$in`, `$gt`, ...) are unchanged.

Tests cover the check directly (including operators nested under a field and inside an `$or` array) plus an integration test that saves a `$where` constraint (rejected) and a normal constraint (accepted) against a real MongoDB.

### DIFF
--- a/.changeset/mongo-unique-operator-guard.md
+++ b/.changeset/mongo-unique-operator-guard.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Reject server-side-evaluation operators (`$where`, `$function`, `$expr`, `$accumulator`) in `unique()` constraints. They have no use in a uniqueness key and could allow query injection when the key is built from untrusted input. Ordinary equality and comparison constraints are unchanged.
--- a/packages/mongo-backend/src/MongoJobRepository.ts
+++ b/packages/mongo-backend/src/MongoJobRepository.ts
@@ -51,6 +51,37 @@ function isPlainObject(value: unknown): value is Record<string, unknown> {
 }
 
 /**
+ * MongoDB query operators that evaluate arbitrary server-side code or expressions.
+ * They have no legitimate use in a uniqueness key and turn a caller-controlled
+ * `unique` object into a NoSQL-injection vector (server-side JS execution, denial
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified: the recursive key-guard correctly bans server-side-eval operators before query construction while preserving the documented unique() filter contract — the right companion to #1728 on the mongo side. Rebased onto main (clean), full suite green locally (259 passed, notably now also under apiStrict from #1785).

simllll (approved):
Verified: the recursive key-guard correctly bans server-side-eval operators before query construction while preserving the documented unique() filter contract — the right companion to #1728 on the mongo side. Rebased onto main (clean), full suite green locally (259 passed, notably now also under apiStrict from #1785).

PR #1733, opened 2026-06-15
cap trailing debounce nextRunAt at the maxWait deadline

Noethix55555:
`maxWait` is documented to guarantee a trailing-debounced job runs within `maxWait` even if saves keep coming. Each save reschedules `nextRunAt` to `now + delay` without bounding it by `debounceStartedAt + maxWait`. So a save shortly before the deadline pushes `nextRunAt` past the deadline, and if saves then stop, the job runs late (up to `delay` after `maxWait`).

Example: `delay = 2000`, `maxWait = 2000`. First save at T0 schedules T2000. A second save at T500 reschedules to T2500, which is past the T2000 deadline.

Fix caps `nextRunAt` at `debounceStartedAt + maxWait` in all three backends (Mongo, Postgres, Redis). The existing maxWait-exceeded branch (force immediate) is unchanged.

Added `should not reschedule past the maxWait deadline` to the shared debounce suite, so it runs against all backends. It fails on the current code and passes with the change. Verified the full maxWait suite on Mongo (memory-server), Postgres, and Redis.

### DIFF
--- a/.changeset/debounce-maxwait-cap.md
+++ b/.changeset/debounce-maxwait-cap.md
@@ -1,0 +1,7 @@
+---
+'@agendajs/mongo-backend': patch
+'@agendajs/postgres-backend': patch
+'@agendajs/redis-backend': patch
+---
+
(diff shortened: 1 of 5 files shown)


fu050409 (approved):
It looks good to me

PR #1735, opened 2026-06-15
preserve startDate, endDate and skipDays in the Mongo backend

Noethix55555:
`saveJob` persists `startDate`, `endDate` and `skipDays` (they ride along in the spread props), but `computeJobObj` — the single function that maps a stored Mongo document back to `JobParameters`, used by `getJobById`, `queryJobs`, `getNextJobToRun`, `lockJob`, and `saveJob`'s own return value — never copied them back. So every job read from the database came back with those constraints `undefined`.

The observable effect: when the processor picks up a repeating job and `Job.run()` recomputes `nextRunAt` via `computeFromInterval`, the constraints are gone, so the job can be scheduled onto days it was told to skip and outside its start/end window.

Fix maps the three fields in `computeJobObj`. Added a test that saves a job with all three and asserts they survive `getJobById` and `queryJobs`; it fails on the current code and passes with the change.

### DIFF
--- a/.changeset/mongo-preserve-date-constraints.md
+++ b/.changeset/mongo-preserve-date-constraints.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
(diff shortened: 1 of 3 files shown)


simllll:
Closing as superseded: the same fix (mapping the three fields in `computeJobObj`) was merged today via #1787, with a regression test covering all read paths. Your diagnosis here was spot-on and predates it by a month — sorry it sat unreviewed. The fix ships with the next @agendajs/mongo-backend patch release.

PR #1739, opened 2026-06-16
preserve startDate, endDate and skipDays in the Redis backend

Noethix55555:
The Redis backend never persists or reads back a job's `startDate`, `endDate` or `skipDays`. `jobToHash` does not write them, the `RedisJobData` hash type has no slots for them, and `hashToJob` does not read them. So a job saved with these date constraints loses them on every reload (`getJobById`, `getNextJobToRun`, re-fetch after restart), silently disabling its start/end window and skipped days.

`JobParameters` defines these fields (`packages/agenda/src/types/JobParameters.ts`) and the scheduler uses them in `utils/dateConstraints.ts` and `utils/nextRunAt.ts`.

Fix serializes the three fields in `jobToHash` and deserializes them in `hashToJob` (dates via ISO strings, `skipDays` via JSON), and adds the slots to `RedisJobData`. Added a round-trip test to the Redis-specific suite; it fails on the current code and passes with the change.

### DIFF
--- a/.changeset/redis-preserve-date-constraints.md
+++ b/.changeset/redis-preserve-date-constraints.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/redis-backend': patch
+---
+
+Persist and restore `startDate`, `endDate` and `skipDays` in the Redis backend. These fields were never written by `jobToHash` nor read by `hashToJob`, so a job saved with date constraints lost them on every reload, silently disabling its start/end window and skipped days.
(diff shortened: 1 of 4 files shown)


simllll:
Closing as superseded: #1789 (merged today) covers this — `jobToHash`/`hashToJob` serialization plus the three partial-update paths in `saveJob`, with a round-trip test through all read paths. Your diagnosis here was correct and a month ahead — sorry it sat unreviewed. Ships with the next @agendajs/redis-backend patch release.

PR #1741, opened 2026-06-16
preserve startDate, endDate and skipDays in the Postgres backend

Noethix55555:
The Postgres backend never persists or reads back a job's `startDate`, `endDate` or `skipDays`. They are not columns in `schema.ts`, not written by any `saveJob` path, and not read by `rowToJob`, so a job saved with these date constraints loses them entirely. The scheduler uses these fields (`utils/dateConstraints.ts`, `utils/nextRunAt.ts`) to gate when a job runs, so the constraints are silently disabled.

This adds `start_date`, `end_date` and `skip_days` columns to the schema and `PostgresJobRow`, writes them in every save path (default insert, single-type upsert including its `ON CONFLICT DO UPDATE`, debounce insert/update, and the by-id update), and maps them in `rowToJob`.

Added a mock-pool test asserting the constraints are written on both the default insert and the single-type upsert and read back via `rowToJob`; it fails on the current code and passes with the change.

### DIFF
--- a/.changeset/postgres-preserve-date-constraints.md
+++ b/.changeset/postgres-preserve-date-constraints.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/postgres-backend': patch
+---
+
+Persist and restore `startDate`, `endDate` and `skipDays` in the Postgres backend. These were not columns in the schema, were not written by `saveJob`, and were not read by `rowToJob`, so a job saved with date constraints lost them entirely, silently disabling its start/end window and skipped days. Adds the columns and wires them through every save path and the row mapping.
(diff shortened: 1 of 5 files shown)


simllll:
Closing as superseded: #1789 (merged today) covers this — the three columns, all five save paths, `rowToJob` mapping, plus an `ALTER TABLE ... ADD COLUMN IF NOT EXISTS` migration so existing installations upgrade automatically on connect. Your diagnosis here was correct and a month ahead — sorry it sat unreviewed. Ships with the next @agendajs/postgres-backend patch release.

PR #1743, opened 2026-06-16
do not skip a cron occurrence when nextRunAt is ahead of the next tick

Noethix55555:
`computeFromInterval` re-rolls the computed cron tick when `nextDate <= previousNextRunAt` (the stored `attrs.nextRunAt`). But `lastRunAt` and the stored `nextRunAt` are unrelated quantities: `nextRunAt` is the scheduled-but-not-yet-fired time. When it is further in the future than the true next tick after `lastRunAt`, the guard bumps `currentDate` forward and recomputes, skipping the legitimate occurrence.

Example: interval `*/13 * * * *`, `lastRunAt = 2024-01-23T20:12:59Z`, `nextRunAt = 2024-01-23T21:52:49Z`. The next tick should be `20:13:00`, but the code returns `20:26:00`.

The guard's documented intent is only to handle cron-parser returning the same instant as the last run, so it now re-rolls only when the tick is not strictly after `lastRunAt`. Added a regression test; it fails on the current code and passes with the change. Full `unit.test.ts` (88) passes.

### DIFF
--- a/.changeset/fix-cron-skip-occurrence.md
+++ b/.changeset/fix-cron-skip-occurrence.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Fix recurring cron jobs skipping a valid occurrence. `computeFromInterval` compared the next cron tick against the stored `nextRunAt`, so when `nextRunAt` was further in the future than the next tick after `lastRunAt`, it bumped past and skipped the legitimate occurrence. It now only re-rolls when the tick is not strictly after `lastRunAt`.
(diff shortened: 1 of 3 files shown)


simllll (approved):
Verified the analysis: the old `<= previousNextRunAt` re-roll fired only once, so it never enforced monotonicity vs the stored nextRunAt — it just skipped exactly one legitimate occurrence in the early-run/stale-nextRunAt case (your 20:26-still-≤-21:52 example proves it wasn't intentional). The new `nextDate <= lastRun` condition implements exactly what the original comment describes. Rebased onto main; agenda (174) and mongo-backend (253) suites green locally, lint clean.

PR #1750, opened 2026-06-17
fix(agenda): unlock future-dated job freed by the processor

### LINKED ISSUE #1749: Recurring job left locked by a crashed worker can be starved forever
A recurring job whose lock outlived its run (for example the worker crashed or was killed mid-run) can stop running permanently.

When the job is picked up again, `getNextJobToRun` matches it through the expired-lock recovery branch (a stale `lockedAt`), regardless of `nextRunAt`. If its `nextRunAt` is further out than `processEvery`, `JobProcessor` takes the "too far away" branch in `packages/agenda/src/JobProcessor.ts`:

```ts
this.lockedJobs.splice(lockedJobIndex, 1);
this.updateStatus(job.attrs.name, 'locked', -1);
```

It removes the job from the local locked list but never calls `unlockJob`, so `lockedAt` stays set in the database. On the next scan the job is matched again through the same expired-lock branch, re-locked (refreshing `lockedAt`), and freed again. With the default `lockLifetime` (10 min) much larger than `processEvery` (5 s) the lock never expires and the job never reaches its `nextRunAt`.

`jobQueueFilling` already calls `unlockJob` when it frees a job; the "too far away" branch should do the same. PR follows.

Noethix55555:
Fixes #1749.

When `findAndLockNextJob` locked a job via the expired-lock recovery path while its `nextRunAt` was still in the future, the "too far away" branch removed it from the local locked list but never released the database lock. The job was re-locked on every scan, refreshing `lockedAt` so the lock never expired, and it never reached its run time. This is the state of a recurring job left locked by a crashed or restarted worker.

The branch now calls `this.agenda.db.unlockJob(job.attrs)`, matching what `jobQueueFilling` already does when it frees a job.

Added a regression test using a minimal in-memory backend that seeds a future-dated job with a stale lock and asserts it runs near its `nextRunAt`. It fails on the previous code (the job never runs) and passes with the fix. Lint and the existing agenda tests pass.

### DIFF
--- a/.changeset/fix-jobprocessor-stale-lock-future-job.md
+++ b/.changeset/fix-jobprocessor-stale-lock-future-job.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Release the database lock when the processor frees a job whose `nextRunAt` is too far in the future. Previously such a job, when picked up via the expired-lock recovery path (for example a recurring job left locked by a crashed worker), was removed from the local locked list but never unlocked in the database, so it was re-locked on every scan and never reached its run time.
--- a/packages/agenda/src/JobProcessor.ts
+++ b/packages/agenda/src/JobProcessor.ts
@@ -516,6 +516,11 @@ export class JobProcessor {
 
 						this.lockedJobs.splice(lockedJobIndex, 1);
 						this.updateStatus(job.attrs.name, 'locked', -1);
+						// Release the database lock as well. Otherwise a job that was
+						// locked via the expired-lock recovery path while its nextRunAt
+						// is still in the future gets re-locked on every scan and never
+						// reaches its run time.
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified the leak: getNextJobToRun's expired-lock recovery branch has no nextRunAt condition, so a future-dated job left locked by a dead worker is re-locked with a fresh lockedAt, freed only locally in the too-far-away branch, and the DB lock cycles forever — the job can never run on time. The fix matches jobQueueFilling's existing release pattern and composes cleanly with #1782 (the freed job never reaches runningJobs, and after the splice it wasn't in stop()'s unlock set anyway). The un-awaited call matches the existing idiom at the jobQueueFilling site; adding .catch to both is follow-up material. Rebased onto main, agenda suite green (incl. the new stale-lock regression test), lint clean.

PR #1754, opened 2026-06-17
fix(agenda): treat numeric repeatInterval as milliseconds

### LINKED ISSUE #1753: Numeric millisecond repeatInterval silently stops recurring jobs after one run
## Description

The public API allows scheduling a recurring job with a numeric millisecond interval:

```ts
await agenda.every(5000, 'job'); // number ms
```

The Mongo backend preserves the number, but the Redis and Postgres backends persist `repeatInterval` as a string (Redis does `String(repeatInterval)`, Postgres uses a `VARCHAR` column). On the next run, the interval is read back as the string `5000`.

In `computeFromInterval` (`packages/agenda/src/utils/nextRunAt.ts`), a numeric string like `5000` is:

- not a valid cron expression (throws), and
- not a valid human interval (`humanInterval(5000)` is `NaN`),

so the function throws `failed to calculate nextRunAt due to invalid repeat interval`, `nextRunAt` becomes `null`, and the recurring job silently stops after its first run.

## Expected

A numeric millisecond interval should keep recurring regardless of whether the backend stores it as a number or a string.

## Fix

In `computeFromInterval`, before the cron/human-interval branches, treat a numeric `repeatInterval` (or a string that is purely numeric) as milliseconds: `nextRunAt = lastRunAt ? new Date(lastRun + n) : new Date(lastRun)`, then apply the existing date-constraint logic. This is a centralized fix that benefits all backends.

Noethix55555:
Fixes #1753

## Problem

`agenda.every(5000, 'job')` accepts a numeric millisecond interval. Mongo preserves the number, but the Redis and Postgres backends persist `repeatInterval` as a string, so it is read back as `5000`. In `computeFromInterval`, a numeric string is neither a valid cron expression nor a valid human interval, so the function throws, `nextRunAt` becomes `null`, and the recurring job silently stops after one run.

## Fix

In `computeFromInterval`, before the cron/human-interval branches, treat a numeric `repeatInterval` (or a string that is purely numeric) as milliseconds:

```ts
nextRunAt = attrs.lastRunAt
  ? new Date(lastRun.valueOf() + numericInterval)
  : new Date(lastRun.valueOf());
```

The existing date-constraint logic still applies. This is centralized so all backends benefit.

## Tests

Added `test/nextRunAt.test.ts` covering numeric-number and numeric-string intervals (the string case fails on the previous code with the thrown `invalid repeat interval` error), plus regression coverage that cron and human intervals still work.

### DIFF
--- a/.changeset/numeric-ms-repeat-interval.md
+++ b/.changeset/numeric-ms-repeat-interval.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Treat a numeric (or purely-numeric string) repeatInterval as milliseconds in computeFromInterval, so recurring jobs scheduled with `agenda.every(5000, 'job')` keep running on backends that persist the interval as a string.
--- a/packages/agenda/src/utils/nextRunAt.ts
+++ b/packages/agenda/src/utils/nextRunAt.ts
@@ -33,7 +33,25 @@ export const computeFromInterval = (attrs: JobParameters<unknown>): Date | null 
 	let nextRunAt: Date | null = null;
 
 	let error;
-	if (typeof attrs.repeatInterval === 'string') {
+
+	// A numeric repeatInterval (or a string that is purely numeric) is treated as
+	// milliseconds. The public API allows `agenda.every(5000, 'job')`, and some
+	// backends persist the interval as a string (e.g. "5000"), which is neither a
+	// valid cron expression nor a human interval and would otherwise throw.
+	const numericInterval =
+		typeof attrs.repeatInterval === 'number'
+			? attrs.repeatInterval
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified empirically on main: string '5000' (what postgres/redis persist) throws `invalid repeat interval` while number 5000 works via humanInterval's number pass-through — so recurring numeric-interval jobs silently die after one run on those backends. The numeric branch exactly mirrors the existing human-interval semantics including the `!lastRunAt` immediate-first-run shape and skipImmediate compatibility, so number-typed intervals see zero behavior change. Rebased onto current main, agenda suite (183) green locally, full CI matrix green.

Issue #1712, opened 2026-05-16
Agendash crashing due to large collection of 14k records


gregg-cbs:
## Description
I am seeing the following error for the "http://localhost:3001/dash/api?limit=50" route:
Executor error during find command: utickets.queue :: caused by :: Sort exceeded memory limit of 33554432 bytes, but did not opt in to external sorting.

If i connect to my other mongo database that has 1000 records, everything works fine.

## Steps to Reproduce

1. Run agendash
2. Have a mongo database with big collection of queue items around 14k
3. Load the dashboard and notice the error

## Expected Behavior
Agendash to work no matter how may records in the job collection

## Environment
- Node.js version: 24
- Agenda version: 6.2.5
- Backend (MongoDB/PostgreSQL/Redis): 4/0/2
- OS: Windows

gregg-cbs:
I suspect its heavily unoptimized if its crashing on a collection of 14 000 records, which is very small amount of records for any database.

gregg-cbs:
Main issue is here:
https://github.com/agenda/agenda/issues/1713

gregg-cbs:
When I apply the patch from https://github.com/agenda/agenda/issues/1713 the dashboard runs.

Its worth noting the dashboard is extremely slow and so it seems there are lots of unoptimized queries that need to fixed.

simllll:
Fixed by #1755 (merged): queryJobs now pushes skip/limit to the database when no state filter is requested (with a parallel countDocuments for the total), instead of materializing the entire collection — which is what crashed agendash on your 14k-job collection. Note #1714 (index recommendations) stays open as a separate enhancement. Ships with the next @agendajs/mongo-backend patch release.

gregg-cbs:
Nice work! I have been clearing out the collection to get around this for the time being.

Sorry i couldnt contribute more, i have been drowning in work! Thanks to those who made this happen!

PR #1755, opened 2026-06-17
fix(mongo-backend): paginate queryJobs at the database

### LINKED ISSUE #1713: [Critical]: Agenda MongoJobRepository.queryJobs has no limit and causing a memory error
## Description

MongoJobRepository.queryJobs has no limit and is dumping all documents from the queue collection into memory or causing the find() to error in mongo due to trying to sort too many records in the database.

The code offender is this line in MongoJobRepository.queryJobs:
```ts
  // Fetch jobs
  const allJobs = await this.collection
  .find(query)
  .sort(this.toMongoSort(sort))
  .toArray()
```

Note that is not using the `limit` param, and the limit param could be 0 which is a bug too.

This could crash production servers or cause jobs to stop processing due to mongo not returning any records because of it erroring when trying to sort and process to many records.

## Steps to Reproduce

1. Have a mongo db with a queue collection of 14k records
2. Run agendash and notice the app errors due to mongo out of memory error (as seen here https://github.com/agenda/agenda/issues/1712)
3. Potentially the same error for agenda processor because the limit is never used

## Expected Behavior
Agenda to not crash or have data memory issues on any amount of queue record data.

## Actual Behavior
Dumping entire queue collection into memory and asking mongo to sort an entire collection of documents - causing errors.

## Environment
- Node.js version: 24
- Agenda version: 6.2.5
- Backend (MongoDB/PostgreSQL/Redis): 4.0.2
- OS: Windows

Noethix55555:
Fixes #1713 (also addresses #1712)

## Problem

`MongoJobRepository.queryJobs` fetched the entire matching collection with `find(query).sort(...).toArray()` and then paginated in JavaScript. Because job `state` is computed in application code, everything was pulled into memory even for a simple paginated query. On large collections this causes out-of-memory errors.

## Fix

When no `state` filter is requested, the state is derived purely from stored fields and cannot exclude documents, so pagination can be pushed to the database:

- `cursor.skip(skip).limit(limit)` for the page
- `collection.countDocuments(query)` for the total (run in parallel)

When a `state` filter IS present, the previous fetch-filter-slice behavior is preserved, since state cannot be expressed as a Mongo query. The `limit: 0` ("no limit") default is also preserved.

## Tests

Added a mongo-backend test that inserts 200 jobs and spies on the repository cursor's `toArray`, asserting that a state-free `queryJobs({ name, limit: 10 })` materializes only 10 documents while `total` reflects the full count. The test fails on the previous implementation (materializes all 200) and passes with the fix.

### DIFF
--- a/.changeset/mongo-query-db-pagination.md
+++ b/.changeset/mongo-query-db-pagination.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Push skip/limit to the database in queryJobs when no state filter is requested, instead of loading the entire matching collection into memory. Prevents out-of-memory on large collections (#1713, #1712).
--- a/packages/mongo-backend/src/MongoJobRepository.ts
+++ b/packages/mongo-backend/src/MongoJobRepository.ts
@@ -272,10 +272,41 @@ export class MongoJobRepository implements JobRepository {
 			query.disabled = { $ne: true };
 		}
 
-		// Fetch jobs
-		const allJobs = await this.collection.find(query).sort(this.toMongoSort(sort)).toArray();
+		const mongoSort = this.toMongoSort(sort);
 
-		// Compute states and filter by state if specified
+		// When there is no state filter, the state is computed purely from the
+		// stored fields and cannot exclude documents, so pagination can be pushed
+		// to the database. This avoids materializing the entire matching
+		// collection into memory (which can OOM on large collections).
+		if (!state) {
+			let cursor = this.collection.find(query).sort(mongoSort);
+			if (skip > 0) {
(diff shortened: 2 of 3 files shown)


simllll (approved):
Verified the equivalence claim: without a state filter the old predicate passed every document (`if (!state) return true`), so DB-side skip/limit + parallel countDocuments returns identical results — total, page contents, and the limit:0 no-limit default all preserved; state-filter path untouched. The cursor-spy test design (failing-first, asserts only the page materializes) is solid. Rebased onto current main (import-line conflict with the #1783/#1788 tests), suite green locally (260 passed), lint warning-free after adding house-style eslint-disable comments.

PR #1759, opened 2026-06-17
fix(agenda): prevent concurrent start() from leaking a second JobProcessor

### LINKED ISSUE #1758: Concurrent start() calls leak a second JobProcessor and double-connect the notification channel
## Description

`Agenda.start()` guards against re-entry with `if (this.jobProcessor) return;`, but that check sits *after* two `await` points (`await this.ready` and `await this.notificationChannel.connect()`). Two concurrent `start()` calls both pass the guard before either assigns `this.jobProcessor`, so two `JobProcessor` instances get created.

## Impact

- The first JobProcessor's `setInterval` is orphaned (leak) — `stop()` only tears down the one stored in `this.jobProcessor`.
- Jobs can be executed twice.
- The notification channel's `connect()` runs twice.

## Reproduction

With a backend whose `connect()` is deferred (~30ms) and a notification channel that counts `connect()` calls:

```ts
await Promise.all([agenda.start(), agenda.start()]);
// channel.connectCount === 2  (expected 1)
```

## Fix

Add a synchronous in-flight guard: cache the start promise and return it for re-entrant calls so only one JobProcessor is ever created and the channel connects once.

Noethix55555:
Fixes #1758

## Problem

`Agenda.start()` checks `if (this.jobProcessor) return;` only after `await this.ready` and `await this.notificationChannel.connect()`. Two concurrent `start()` calls both clear the guard before either assigns `this.jobProcessor`, creating two JobProcessors. This leaks the first processor's `setInterval` (since `stop()` only tears down one), causes duplicate job execution, and connects the notification channel twice.

## Fix

Cache the in-flight start promise in a private `startPromise` field. Re-entrant calls await the existing promise instead of starting again, so exactly one JobProcessor is created and the channel connects once.

## Test

Added a unit test (`test/unit.test.ts`) with a deferred-connect backend and a connect-counting channel: `await Promise.all([agenda.start(), agenda.start()])` results in exactly one channel connect and one live processor; after `stop()` the processor is gone and the channel is disconnected. Verified the test fails on the current code (connectCount === 2) and passes with the fix.

### DIFF
--- a/.changeset/start-double-start-race.md
+++ b/.changeset/start-double-start-race.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Fix start() race where concurrent calls leaked a second JobProcessor and connected the notification channel twice
--- a/packages/agenda/src/index.ts
+++ b/packages/agenda/src/index.ts
@@ -238,6 +238,8 @@ export class Agenda extends EventEmitter {
 
 	private jobProcessor?: JobProcessor;
 
+	private startPromise?: Promise<void>;
+
 	readonly ready: Promise<void>;
 
 	isActiveJobProcessor(): boolean {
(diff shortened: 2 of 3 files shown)


simllll (approved):
Race verified on current main: the jobProcessor guard sits after `await this.ready`, so concurrent callers both pass it — double processor, double channel connect, leaked interval. The cached in-flight promise with `finally` cleanup is the right fix (failed starts can retry; both callers observe the same rejection). Rebased onto main, agenda suite green locally (174 passed), lint clean.

PR #1777, opened 2026-06-17
fix(test): send fork child error message over IPC instead of empty object

### LINKED ISSUE #1776: Fork helper sends empty object instead of error message over IPC, corrupting failReason
## Problem

The reference fork helpers (`packages/{mongo,postgres,redis}-backend/test/helpers/forkHelper.ts`, all identical at ~line 49) report a child error to the parent via:

```ts
process.send(JSON.stringify(err));
```

But `JSON.stringify(error)` of an `Error` instance is `"{}"` (Error's own enumerable properties are empty). The real `message` is lost. The parent then parses and persists `failReason` as an empty object, which violates the typed contract (`failReason` is typed as `string`).

## Fix

```ts
process.send(err instanceof Error ? err.message : JSON.stringify(err));
```

so the real message is sent for `Error` instances (and non-Error values still fall back to JSON).

## Verification

The existing fork failure-mode test only asserted `failReason` is not undefined, so it didn't catch this. Strengthened to assert `typeof failReason === 'string'` (and non-empty, not `'{}'`). On a real forked run with mongodb-memory-server:
- Before the fix: `typeof failReason === 'object'` (the persisted `{}`).
- After the fix: `typeof failReason === 'string'` (the real message).

These are test/reference helpers, so no published-package changeset is needed.

Noethix55555:
Fixes #1776

## Problem

The reference fork helpers (`packages/{mongo,postgres,redis}-backend/test/helpers/forkHelper.ts`) send a child error to the parent with `process.send(JSON.stringify(err))`. `JSON.stringify(Error)` is `"{}"`, so the real `message` is lost and the parent persists `failReason` as an empty object (violating the `failReason: string` type).

## Fix

`process.send(err instanceof Error ? err.message : JSON.stringify(err))` in all three helpers.

## Test

Strengthened the shared `should handle job failure in fork mode` test to assert `failReason` is a non-empty string (not the `'{}'` artifact). Verified with a real forked run on mongodb-memory-server:
- HEAD (old helper): `typeof failReason === 'object'` -> test fails.
- With fix: `typeof failReason === 'string'` -> test passes.

Note: the fork happy-path import currently hits a separate, already-known Windows `file://` URL issue in this environment, so the specific thrown message (`intended error :-)`) cannot surface locally on Windows. The error-serialization fix is nonetheless verified end-to-end: the surfaced error message is now a real string rather than `{}`. The assertions are intentionally message-agnostic so they hold across environments.

No changeset: these are test/reference helpers, not part of a published package.

### DIFF
--- a/packages/agenda/test/shared/agenda-test-suite.ts
+++ b/packages/agenda/test/shared/agenda-test-suite.ts
@@ -1954,6 +1954,13 @@ export function agendaTestSuite(config: AgendaTestConfig): void {
 					const jobDataFinished = await backend.repository.getJobById(job.attrs._id!);
 					expect(jobDataFinished?.lastFinishedAt).toBeDefined();
 					expect(jobDataFinished?.failReason).not.toBeUndefined();
+					// failReason must be the real error message string, not a serialized
+					// Error object. JSON.stringify(error) yields "{}", which would be
+					// persisted as an empty object (losing the message and violating the
+					// `failReason: string` type). The fork helper sends `err.message`.
+					expect(typeof jobDataFinished?.failReason).toBe('string');
+					expect((jobDataFinished?.failReason as string).length).toBeGreaterThan(0);
+					expect(jobDataFinished?.failReason).not.toBe('{}');
 					expect(jobDataFinished?.failCount).toBe(1);
 
 					await agendaFork.stop();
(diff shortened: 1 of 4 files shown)


simllll (approved):
Verified both sides of the IPC contract: `JSON.stringify(Error)` is indeed `"{}"` (non-enumerable props), and the parent (Job.ts:679-687) has a try/catch fallback that uses unparseable strings raw — so `err.message` flows into `failReason` as a proper string, while the old flow persisted an empty object violating the `failReason: string` type. No-changeset is correct (test/reference helpers only). Rebased onto current main, mongo suite green incl. the strengthened fork-failure assertion.

PR #1780, opened 2026-06-17
fix(agenda): honor repeatTimezone when resolving repeatAt

### LINKED ISSUE #1779: repeatAt ignores repeatTimezone (wrong wall-clock time on non-local servers)
## Description

When a job is scheduled with both `repeatAt` and `repeatTimezone`, the `repeatTimezone` is ignored.

`computeFromRepeatAt` in `packages/agenda/src/utils/nextRunAt.ts` resolves the `repeatAt` string via `date.js`, which only parses the time-of-day in the **server's local timezone** (no timezone argument is passed). As a result, a job with:

```ts
{ repeatAt: '3:30pm', repeatTimezone: 'Asia/Tokyo' }
```

fires at 3:30pm in the **server's** timezone rather than 3:30pm in Tokyo.

## Expected

The cron interval path (`computeFromInterval`) already passes `tz: attrs.repeatTimezone` to cron-parser, so cron-based repeats honor the timezone. The `repeatAt` path should behave consistently.

## Impact

On any server not running in the configured timezone, `repeatAt` jobs run at the wrong wall-clock time. The discrepancy equals the offset between the server timezone and `repeatTimezone`.

Noethix55555:
Fixes #1779

## Problem

`computeFromRepeatAt` (`packages/agenda/src/utils/nextRunAt.ts`) parses the `repeatAt` time-of-day with `date.js`, which resolves only against the server's local timezone. `attrs.repeatTimezone` was ignored, so a job with `repeatAt: '3:30pm'` and `repeatTimezone: 'Asia/Tokyo'` fired at 3:30pm in the server's zone rather than in Tokyo.

The cron interval path (`computeFromInterval`) already passes `tz: attrs.repeatTimezone` to cron-parser, so this was a cross-path inconsistency.

## Fix

When `repeatTimezone` is set, reinterpret the parsed wall-clock time in the target zone using Luxon (already imported):

```ts
nextRunAt = DateTime.fromJSDate(nextRunAt)
  .setZone(attrs.repeatTimezone, { keepLocalTime: true })
  .toJSDate();
```

Applied to both the main and "tomorrow at ..." branches, before the date-constraint logic. Behavior is unchanged when `repeatTimezone` is unset.

## Tests

Added unit tests in `packages/agenda/test/unit.test.ts`:
- `repeatAt: '3:30pm'` with `Asia/Tokyo` vs `America/New_York` must differ by more than 12h (the zone offset). Fails on current HEAD (both collapse to the same instant), passes with the fix.
- With no `repeatTimezone`, the result still matches a plain local parse.

`npx eslint` passes on the changed files; the full `unit.test.ts` suite passes.

### DIFF
--- a/.changeset/repeatat-timezone.md
+++ b/.changeset/repeatat-timezone.md
@@ -1,0 +1,9 @@
+---
+'agenda': patch
+---
+
+Honor `repeatTimezone` when computing `nextRunAt` from `repeatAt`. Previously the
+`repeatAt` time-of-day was resolved only in the server's local timezone, so a job
+with `repeatAt: '3:30pm'` and `repeatTimezone: 'Asia/Tokyo'` would fire at the
+wrong wall-clock time on a non-Tokyo server. The parsed time is now reinterpreted
+in the configured timezone, matching the cron interval path.
--- a/packages/agenda/src/utils/nextRunAt.ts
+++ b/packages/agenda/src/utils/nextRunAt.ts
@@ -120,6 +120,16 @@ export function computeFromRepeatAt(attrs: JobParameters<unknown>): Date | null 
 		nextRunAt = date(repeatAt);
 	}
 
+	// `date.js` resolves `repeatAt` against the server's local timezone only. When a
+	// `repeatTimezone` is configured, reinterpret the parsed wall-clock time as if it
+	// occurred in that zone so the job fires at the intended local time regardless of
+	// where the process runs (mirrors the `tz` option used by the cron interval path).
+	if (attrs.repeatTimezone) {
+		nextRunAt = DateTime.fromJSDate(nextRunAt)
+			.setZone(attrs.repeatTimezone, { keepLocalTime: true })
+			.toJSDate();
+	}
+
 	// Apply date constraints (startDate, endDate, skipDays)
 	if (attrs.startDate || attrs.endDate || attrs.skipDays) {
 		const constrainedDate = applyAllDateConstraints(nextRunAt, {
--- a/packages/agenda/test/unit.test.ts
+++ b/packages/agenda/test/unit.test.ts
@@ -7,6 +7,7 @@ import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { Agenda, Job, toJobId } from '../src/index.js';
 import type { AgendaBackend, JobRepository, JobParameters, JobLogger, JobLogEntry, JobLogQuery, JobLogQueryResult } from '../src/index.js';
 import { computeJobState } from '../src/types/JobQuery.js';
+import { computeFromRepeatAt } from '../src/utils/nextRunAt.js';
 
 /**
  * Minimal mock repository that satisfies the interface without real storage.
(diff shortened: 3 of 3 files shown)


simllll (approved):

simllll (changes requested):
Thanks for this — the diagnosis is correct and the direction (reinterpret `repeatAt` in `repeatTimezone`, mirroring the cron path) is exactly right. But the implementation misses that **`date.js` is future-biased in server-local time**: a time-of-day that already passed rolls to *tomorrow*. The `keepLocalTime` shift silently destroys that guarantee, which breaks in both directions. Both cases reproduced locally against your branch:

### 1. Zones east of the server → past `nextRunAt` → immediate re-fire loop

At 18:24 CEST (server local), `repeatAt: '11:30pm'` with `repeatTimezone: 'Asia/Tokyo'`:

```
date('11:30pm')                  -> Mon Jul 20 23:30 CEST   (future locally ✓)
.setZone('Asia/Tokyo',
         { keepLocalTime: true }) -> Mon Jul 20 16:30 CEST   (≈2h in the PAST ✗)
```

`computeNextRunAt()` has no past-guard, so a past `nextRunAt` makes the job fire immediately — and each recompute yields the same past instant, producing a tight execution loop that lasts up to the zone-offset in hours, every day.

### 2. Zones west of the server → silently skips today's occurrence

At 18:24 CEST, `repeatAt: '3:30pm'` with `America/New_York`: today's 3:30pm NY is still ~3h in the future, but because date.js already rolled the *local* parse to tomorrow, the fix schedules **tomorrow's** 3:30pm NY — a day late.

### 3. The "unchanged behavior" test is time-of-day dependent

`leaves behavior unchanged when repeatTimezone is unset` asserts the result ≈ *today* 15:30 local — but after 15:30 server time, `date('3:30pm')` returns *tomorrow* 15:30, so the suite fails deterministically whenever CI runs between 15:30 and midnight in the runner's zone. It has only passed so far because runs landed in the morning window.

### Suggested fix

Take the time-of-day from the parse, construct it in the target zone for *today (in that zone)*, and roll forward a day if already past — i.e. re-establish the future-bias contract in the **target** zone:

```ts
if (attrs.repeatTimezone) {
	const parsed = DateTime.fromJSDate(nextRunAt);
	let inZone = DateTime.now().setZone(attrs.repeatTimezone).set({
		hour: parsed.hour,
		minute: parsed.minute,
		second: parsed.second,
		millisecond: 0
	});
	if (inZone.toMillis() <= Date.now()) {
		inZone = inZone.plus({ days: 1 });
	}
	nextRunAt = inZone.toJSDate();
}
```

(Luxon's `plus({ days: 1 })` keeps this DST-safe. Consider also throwing on an invalid `repeatTimezone` — `setZone` returns an invalid DateTime silently, whereas the cron path throws.)

Note the corrected fix also invalidates the Tokyo-vs-NY `> 12h` assertion: with proper day-rolling the gap between the two zones can legitimately be **11h** (24 − 13) depending on when the test runs. I'd suggest freezing time with `vi.setSystemTime` and asserting exact expected instants for all three tests — that removes every time-of-day dependency at once.

Issue #1710, opened 2026-05-13
One-time jobs created with `agenda.now()` are executed repeatedly when `removeOnComplete` is false


derwaldgeist:
### Environment

- `agenda`: 6.2.5
- `@agendajs/mongo-backend`: 4.0.2
- MongoDB driver: 6.9.0
- Node.js: 22.x

### Summary

One-time jobs created with `agenda.now()` are executed repeatedly when `removeOnComplete` is left at its default value (`false`).

After the job completes successfully, the MongoDB document keeps its original `nextRunAt` value in the past. Since the polling query treats `nextRunAt <= now` as runnable, the same job is picked up again on subsequent polls.

### Reproduction

```ts
import { Agenda } from 'agenda';
import { MongoBackend } from '@agendajs/mongo-backend';

const agenda = new Agenda({
  backend: new MongoBackend({
    address: 'mongodb://localhost:27017/agenda-test',
    collection: 'agenda-jobs'
  })
});

agenda.define('heartbeat', async job => {
  console.log('heartbeat', job.attrs._id, new Date());
}, {
  lockLifetime: 10_000
  // removeOnComplete is false by default
});

await agenda.start();
await agenda.now('heartbeat');
```

nextRunAt stays in the past, so the job continues to match the runnable-job query.

### Expected behavior
A successfully completed one-time job should not be executed again.

I would expect one of the following:

* nextRunAt is persisted as null
* the job is disabled
* completed one-time jobs are excluded from the runnable-job query
* or one-time jobs are removed automatically, if that is the intended behavior

### Possible cause
In Job.computeNextRunAt(), one-time jobs appear to set:

`this.attrs.nextRunAt = null;`

But in MongoJobRepository.saveJobState(), nextRunAt is written like this:

`nextRunAt: (job.nextRunAt && new Date(job.nextRunAt)) || undefined`

So null becomes undefined, and the previous nextRunAt value is not cleared in MongoDB.

This appears to leave completed one-time jobs permanently runnable.

### Workaround
Setting removeOnComplete: true avoids the issue:

```ts
agenda.define('heartbeat', handler, {
  removeOnComplete: true
});
```

However, this means completed jobs cannot be retained in MongoDB for dashboard/history purposes.

ciak0:
This could be caused by the `ignoreUndefined` flag of the MongoDB driver.
I am seeing multiple issues with repeating jobs or jobs that aren't removed when using it.

derwaldgeist:
[...]

PR #1783, opened 2026-06-29
fix(mongo-backend): explicit set null values and avoid relying on the ignoreUndefined flag

ciak0:
Made all optional Job keys `null` in the MongoDB repository so that the behaviour is not influenced by the native `ignoreUndefined` flag of the MongoDB driver

### Context
The MongoDB driver has a flag called "ignoreUndefined".
This flag ignores all the `undefined` values that are set in the provided JSON objects when serialized to the MongoDB server, instead of setting them to `null`.

This flag can be set by the consumers of the Agenda library to persist arbitrary objects in the Job data field.
The presence of the `ignoreUndefined` flag makes the Agenda MongoDB repository act in unexpected ways since it tries to save the Job native options by setting them as `undefined`.

This creates all kinds of issues with repeating jobs or jobs that aren't removed from the collection, since all the fields expected to be set to `null` are instead left unchanged in the documents persisted in the Mongo collection.

See https://github.com/agenda/agenda/issues/1710

### Changes
- Change all the optional Job fields to be declared as null in the MongoDB repository
- Change all the write operations to set null as the default value for the optional fields
- Change the mapping from the Mongo document to coerce null values to undefined
- Add test coverage

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

### DIFF
--- a/.changeset/stale-rules-matter.md
+++ b/.changeset/stale-rules-matter.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Extended the saved Job type to have explicit null values instead of undefined and stopped relying on the MongoDb driver "ignoreUndefined" flag to set them
--- a/packages/mongo-backend/src/MongoJobRepository.ts
+++ b/packages/mongo-backend/src/MongoJobRepository.ts
@@ -80,22 +80,22 @@ function computeJobObj<DATA = unknown>(job: WithId<MongoJobDocument>): JobParame
 		nextRunAt: job.nextRunAt,
 		type: job.type,
 		data: job.data as DATA,
-		lockedAt: job.lockedAt || undefined,
-		lastFinishedAt: job.lastFinishedAt || undefined,
-		failedAt: job.failedAt || undefined,
-		failCount: job.failCount || undefined,
-		failReason: job.failReason || undefined,
-		repeatTimezone: job.repeatTimezone || undefined,
-		lastRunAt: job.lastRunAt || undefined,
-		repeatInterval: job.repeatInterval || undefined,
-		repeatAt: job.repeatAt || undefined,
-		disabled: job.disabled || undefined,
-		progress: job.progress || undefined,
-		unique: job.unique || undefined,
-		uniqueOpts: job.uniqueOpts || undefined,
-		lastModifiedBy: job.lastModifiedBy || undefined,
-		fork: job.fork || undefined,
-		debounceStartedAt: job.debounceStartedAt || undefined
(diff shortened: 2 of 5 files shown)


ciak0:
@simllll, a pretty heavy bug that affects the entire Agenda MDB backend functionality via a driver flag.

Do you need an issue for the PR?

simllll:
Is this an issue that exists since the beginning of agenda or just since the V6 rewrite?

I will have a look later. I think the pr is enough, no seperat issue required! Thanks

ciak0:
It might be that exists since the beginning, those lines look pretty old on
git lens!

ciak0:
> Is this an issue that exists since the beginning of agenda or just since the V6 rewrite?
> 
> I will have a look later. I think the pr is enough, no seperat issue required! Thanks

@simllll I have checked v5.0.0 and all the nullable fields were explicitly declared as `lockedAt?: Date | null`, and they were explicitly set to `null` when unlocked. So yes, this is a v6 issue.

ciak0:
@simllll, any news for this? Is there any chance this gets merged and a new release published by September? We're kinda blocked for an upcoming feature. 🙏 👼

simllll:
I will have a look later this month 👍🏼

derwaldgeist:
I want to support this. It effectively breaks keeping finished jobs in the database for me, see #1710.

Since I am using Meteor's included Mongo driver, I can't just set the "ignoreUndefined"-Flag. And I think setting the values to `null` would be the correct way to clear values anyway. Another solution would be to $unset them, but nulling works just fine.

Thanks for the great job solution, btw. Agenda is a perfect match for any Meteor user.

simllll (approved):
Verified the approach: explicit `null` correctly drops completed jobs out of the `nextRunAt: { $lte }` runnable query, and `{ $eq: null }` / `{ $ne: null }` treat explicit null and missing fields identically, so old and new documents coexist without migration. Since the driver default (`ignoreUndefined: false`) already serialized `undefined` as `null`, this is behavior-neutral for default configs and only changes behavior for `ignoreUndefined: true` users — exactly the affected group. Thanks for the fix!

Merging as-is; we'll address some minor hardening in a follow-up:
- restore the defensive `new Date()` wrapping in `saveJobState` (untyped callers passing ISO strings would now persist strings)
- regression test for the original #1710 scenario (completed one-time job via `saveJobState`)
- tabs vs. spaces in the new blocks
- reconsider the public `OptionalKeysToNullable` export

ciak0:
Awesome, thanks!
Waiting for the next release ❤️

PR #1785, opened 2026-07-10
fix(mongo-backend): conform to MongoDB Stable API

ahelmberger:
## Description

Makes `@agendajs/mongo-backend` fully conform to the [MongoDB Stable API](https://www.mongodb.com/docs/manual/reference/stable-api/), so it works with clients configured with `apiStrict: true` and `deprecationErrors: true`.

- Replace the `distinct` command in `MongoJobRepository.getDistinctJobNames()` with an equivalent `$group` aggregation. `distinct` is not part of Stable API V1 and fails with an `APIStrictError` on strictly configured clients; `aggregate` is included.
- Guard against future regressions: all test connections (including the fork-mode worker) now use a shared `testMongoClientOptions` helper that enables `serverApi: { version: '1', strict: true, deprecationErrors: true }`. Since the tests run against a real `mongod` (via `mongodb-memory-server`), any non-conforming command introduced later fails the test suite immediately.

All other commands issued by the package (`find`, `findAndModify`, `insert`, `update`, `delete`, `aggregate`, `createIndexes`, change streams, etc.) were verified to be Stable API V1 conformant.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

### DIFF
--- a/.changeset/mongo-stable-api.md
+++ b/.changeset/mongo-stable-api.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Conform to the MongoDB Stable API: replace the `distinct` command in `getDistinctJobNames()` with a `$group` aggregation, so the backend works with clients configured with `apiStrict: true`
--- a/packages/mongo-backend/src/MongoJobRepository.ts
+++ b/packages/mongo-backend/src/MongoJobRepository.ts
@@ -313,7 +313,12 @@ export class MongoJobRepository implements JobRepository {
 	 * Get all distinct job names
 	 */
 	async getDistinctJobNames(): Promise<string[]> {
-		return this.collection.distinct('name');
+		// Uses $group instead of the distinct command, which is not part of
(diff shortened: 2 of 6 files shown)


simllll:
Verified and rebased. The `$group` replacement is correct (`distinct` is indeed absent from Stable API V1), and the strict-mode test harness is a great permanent regression guard.

Since maintainer edits were enabled, I've rebased your branch onto current main — it conflicted with today's #1783/#1788 changes to `createTestDb` in `mongo-backend.test.ts`. Resolution combines both: `MongoClient.connect(uri, testMongoClientOptions)` (your client-level Stable API options) + `client.db(dbName, options)` (our db-level options param, used by the `ignoreUndefined` regression tests).

Ran the full mongo-backend suite on the rebased branch locally: **253 passed under `apiStrict: true`** — which also confirms everything merged today (#1783, #1787, #1788) is Stable-API-conformant. Lint clean.

ahelmberger:
Thanks, @simllll

PR #1787, opened 2026-07-17
fix(mongo-backend): include startDate, endDate, skipDays in computeJo…

Nagasai-challa:
`computeJobObj` in `MongoJobRepository` does not map `startDate`, `endDate`, or `skipDays` from MongoDB documents to job objects.

These fields are written to MongoDB by `Job.save()` via `toJson()`, and are expected by `computeFromInterval()` in `utils/nextRunAt.js` to enforce date constraints. However, when jobs are loaded from the database (via `getNextJobToRun`, `getJobById`, `queryJobs`, `lockJob`), these fields are dropped.

## Impact

Any job with an `endDate` set will never stop running. After the first execution, `computeNextRunAt()` cannot see the `endDate` constraint because `attrs.endDate` is `undefined`.

## Type of Change

- [x] Bug fix

## Checklist

### DIFF
--- a/.changeset/mongo-date-constraint-fields.md
+++ b/.changeset/mongo-date-constraint-fields.md
@@ -1,0 +1,5 @@
+---
+'@agendajs/mongo-backend': patch
+---
+
+Map `startDate`, `endDate` and `skipDays` from MongoDB documents to job objects so date constraints on repeating jobs survive a reload from the database (previously a repeating job with an `endDate` would keep running forever after its first execution)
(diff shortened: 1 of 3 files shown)


simllll:
Thanks for the fix — verified it's correct: `computeFromInterval()` enforces `startDate`/`endDate`/`skipDays` from `attrs`, and `computeJobObj` (the single read-path mapper used by `getNextJobToRun`/`getJobById`/`queryJobs`/`lockJob`) was dropping them, so a repeating job with an `endDate` would lose the constraint after reload and run forever.

Since "allow edits from maintainers" was enabled, I've pushed to your branch:
- rebased onto `main` (resolved the conflict with #1783/#1788 — the new fields now use the `?? undefined` style consistent with the surrounding code)
- added a regression test covering all read paths
- added the missing changeset

Will merge once CI is green.

simllll (approved):
Verified: `computeFromInterval()` enforces these constraints from `attrs`, and `computeJobObj` is the single read-path mapper — this was the right place for the fix. Regression test covers all read paths (`getJobById`, `queryJobs`, `getNextJobToRun`). Thanks!

Nagasai-challa:
@simllll  Thanks for the quick review and merge! We have a production dependency on this fix, when can we expect the next npm release?

simllll:
Ppbl in a few hours :)

Nagasai-challa:
okay, thank you :)

PR #1794, opened 2026-07-22
fix(agenda): deduplicate skipDays before all-days check

santhiprakash:
## Problem

`skipDays` validation rejected an array like `[0, 0, 1, 2, 3, 4, 5]` as "skipping every weekday" even though a valid weekday (e.g. Saturday) is not actually skipped. Duplicates in the list made `areWeSkippingThisDay` think every day of the week was covered.

## Triage / Root cause

`dateConstraints.ts` compared the index of each day in the original `skipDays` array to `indexOf` to detect duplicates, but then used the whole (non-deduplicated) array for the "all days skipped" check.

## Fix

- Deduplicate `skipDays` with `[...new Set(skipDays)]` before the all-days check.
- Add a regression test in `packages/agenda/test/date-constraints.test.ts`.

## Verification

- `pnpm --filter agenda test` (full suite) passed.
- Added test case `should allow duplicate skipDays values` passes.

## Notes / Risks

No public API change. Previously rejected valid configurations now work; truly invalid configurations (e.g. all seven unique weekdays) are still rejected.

### DIFF
--- a/.changeset/deduplicate-skipdays.md
+++ b/.changeset/deduplicate-skipdays.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Deduplicate `skipDays` before checking whether all weekdays are skipped. Previously an array with duplicate entries (e.g. `[0, 0, 1, 2, 3, 4, 5]`) was rejected as covering every day even though a valid weekday was not skipped.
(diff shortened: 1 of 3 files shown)


simllll (dismissed):
Nice catch (y) thanks

santhiprakash:

PR #1795, opened 2026-07-23
fix(agenda): make isValidDate a true type guard

santhiprakash:
## Problem

`isValidDate` was declared as a TypeScript type predicate `value is Date`, but at runtime it accepted strings and other values that `new Date()` could parse. Code that relied on the type guard could then call `Date` methods on a value that was not actually a `Date` instance, producing runtime errors.

## Triage / Root cause

The implementation only checked `value instanceof Date || !Number.isNaN(new Date(value).getTime())`. A string like `"2026-07-22"` passes `Number.isNaN(new Date(value).getTime())` but is not a `Date`.

## Fix

- Return `false` unless `value instanceof Date` and `Number.isFinite(value.getTime())`.
- Remove the implicit string coercion path.
- Add unit tests in `packages/agenda/test/isValidDate.test.ts` for `Date` instances, invalid dates, strings, numbers, and `null`/`undefined`.

## Verification

- `pnpm --filter agenda exec vitest run packages/agenda/test/isValidDate.test.ts` passed.
- `pnpm --filter agenda test` passed.

## Notes / Risks

This tightens the contract of `isValidDate` to match its type predicate. Call sites that passed non-`Date` values and relied on coercion now get `false` instead of a runtime exception.

### DIFF
--- a/.changeset/isvaliddate-typeguard.md
+++ b/.changeset/isvaliddate-typeguard.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Make `isValidDate` a true type guard by checking for a `Date` instance with a finite `getTime()`. Previously it accepted strings and other values that `new Date()` could parse, while narrowing them to `Date` at the type level, which could cause runtime errors when calling `Date` methods.
--- a/.changeset/isvaliddate-typeguard.md
+++ b/.changeset/isvaliddate-typeguard.md
@@ -1,5 +1,0 @@
----
-'agenda': patch
----
-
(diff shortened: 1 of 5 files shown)


pepepeboom:
Reviewed the production call site and the added tests. The `isValidDate` change itself correctly aligns runtime behavior with the `date is Date` predicate: the only production call site passes `Date | null`, and the new tests cover invalid Date instances and non-Date inputs.

One scope issue before merge: this branch also includes `f587bfa` and the `dateConstraints.ts` / `date-constraints.test.ts` changes already proposed independently in #1794. That duplicates an open fix and creates avoidable merge-order conflicts. Please rebase onto `agenda:main` (or drop `f587bfa` and its two file changes) so #1795 contains only the isValidDate fix and tests. The focused patch then looks ready from my review.

santhiprakash:

PR #1796, opened 2026-07-23
fix(agenda): resolve repeatAt relative to lastRunAt

santhiprakash:
## Description

`computeFromRepeatAt` in `packages/agenda/src/utils/nextRunAt.ts` now resolves `repeatAt` strings relative to `lastRunAt` instead of the current wall-clock time.

Previously, `date(repeatAt)` was computed from `new Date()`. When `lastRunAt` was in the future (for example, a manually scheduled job or a skipped run), the resulting `nextRunAt` could be an instant before `lastRunAt`, causing the job to be scheduled in the past and run repeatedly.

This change:
- Uses `lastRun` as the reference date for every `date.js` call.
- Keeps the existing "tomorrow at" advance when `repeatAt` exactly matches `lastRunAt`.
- Adds regression tests in `packages/agenda/test/nextRunAt.test.ts` covering the past-time case and the exact-match case.

## Relation to open PRs

This is orthogonal to #1780 (repeatTimezone handling in the same function). I checked that PR's diff and it only adds a timezone reinterpretation block after the `nextRunAt` assignment, so whichever merges first should be a straightforward rebase for the other.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Added changeset if needed (`pnpm changeset`)
- [ ] Updated documentation if needed

I did not add a changeset because this is a pure bug fix with no public API change.

### DIFF
--- a/packages/agenda/src/utils/nextRunAt.ts
+++ b/packages/agenda/src/utils/nextRunAt.ts
@@ -118,7 +118,10 @@ export function computeFromRepeatAt(attrs: JobParameters<unknown>): Date | null 
 		throw new Error('repeatAt is required');
 	}
 
-	const nextDate = date(repeatAt).valueOf();
+	// Resolve `repeatAt` from `lastRun`, not from `now`. When `lastRunAt` is in the
+	// future (e.g. a manually scheduled job), `date(repeatAt)` from the current
+	// clock can return an instant before `lastRun`, causing duplicate or missed runs.
+	const nextDate = date(repeatAt, lastRun).valueOf();
 
 	// If you do not specify offset date for below test it will fail for ms
 	const offset = Date.now();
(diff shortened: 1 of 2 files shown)


simllll:
What's the use case? A job can only be rescheduled when it just finished?

santhiprakash:
The main use case is a job whose `lastRunAt` is in the future relative to the wall clock. That happens when a job is scheduled for a future time (for example via `agenda.schedule()` or a manually set `nextRunAt`) and then begins executing around that scheduled time.

If `repeatAt` is resolved from `new Date()` in that moment, a string like `'in 1 minute'` returns `now + 1 min`, which can be *before* `lastRunAt`. The processor then sees a `nextRunAt` in the past and immediately picks the job up again, so it runs repeatedly.

The regression test in this PR reproduces that exact case: `lastRunAt = Date.now() + 120s` and `repeatAt: 'in 1 minute'`. With the old code `nextRunAt` would be `Date.now() + 60s`, i.e. before `lastRunAt`; with the change it is computed relative to `lastRunAt` and is therefore always after it.

It is not limited to "when the job just finished" — it also covers delayed/skipped execution where the stored `lastRunAt` reflects the intended run time rather than the current clock.

santhiprakash:

PR #1797, opened 2026-07-23
fix(agenda): wait for startup before teardown

pepepeboom:
## Description

Fix a race when `stop()` or `drain()` is called while `start()` is still awaiting backend or notification-channel setup. Previously, teardown could return before `jobProcessor` existed, then startup would complete and leave a processor running.

Both teardown methods now await an in-flight `startPromise` before checking and shutting down the processor. Regression tests cover both `stop()` and `drain()`, including cleanup on assertion failure.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Checklist

- [ ] Tests pass (`pnpm test`) - not run locally; upstream CI should verify
- [ ] Linting passes (`pnpm lint`) - not run locally; upstream CI should verify
- [x] Updated documentation if needed (no user-facing documentation change required)

### DIFF
--- a/.changeset/lifecycle-teardown-start-race.md
+++ b/.changeset/lifecycle-teardown-start-race.md
@@ -1,5 +1,5 @@
 ---
-"agenda": patch
+'agenda': patch
 ---
 
 Fix stop() and drain() called during start() from leaving a JobProcessor running.
--- a/.changeset/lifecycle-teardown-start-race.md
+++ b/.changeset/lifecycle-teardown-start-race.md
@@ -1,0 +1,5 @@
+---
+"agenda": patch
+---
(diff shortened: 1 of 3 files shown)


pepepeboom:
Local verification completed on the current head:

- `pnpm exec vitest run test/unit.test.ts`: 91 tests passed
- `pnpm run typecheck` in `packages/agenda`: passed
- `pnpm exec eslint packages/agenda/src/index.ts packages/agenda/test/unit.test.ts`: passed
- `git diff --check`: passed

The full monorepo test/lint workflows remain delegated to upstream CI.

PR #1799, opened 2026-07-23
fix(agenda): clamp near-future setTimeout delays to signed 32-bit max

santhiprakash:
## Problem

`JobProcessor` arms a `setTimeout` for near-future jobs with:

```ts
runIn > MAX_SAFE_32BIT_INTEGER ? MAX_SAFE_32BIT_INTEGER : runIn
```

`MAX_SAFE_32BIT_INTEGER` was `2 ** 31` (2147483648). Node.js treats delays larger than a signed 32-bit integer (`2 ** 31 - 1`) as an overflow and clamps them to `1` ms, so a timer armed with `2 ** 31` fires almost immediately and reprocesses the job in a tight loop.

This is reachable when `processEvery` is configured larger than the job's delay (for example, a human-interval like `processEvery: '30 days'` ≈ 2.59 billion ms). A job due in, say, 10 days would have `runIn` ≈ 864 million ms — larger than `2 ** 31 - 1` and within `processEvery`, so the processor arms the timer instead of releasing the lock.

## Fix

Change the ceiling to `2 ** 31 - 1`, which is the maximum delay Node.js accepts without overflowing.

- `packages/agenda/src/JobProcessor.ts`: `MAX_SAFE_32BIT_INTEGER = 2 ** 31 - 1`
- `packages/agenda/test/jobprocessor-settimeout-overflow.test.ts`: regression test that arms a near-future timer with `runIn` above `2 ** 31` and asserts the clamped delay equals `2 ** 31 - 1`, not `2 ** 31`.
- `.changeset/settimeout-overflow.md`: patch changeset for `agenda`.

## Verification

```bash
cd packages/agenda && bun run test
# Test Files  12 passed (12)
# Tests       187 passed (187)

cd ../..
bun run lint   # clean
bun run typecheck  # clean
```

Also verified the regression test fails when the constant is reverted to `2 ** 31`:

```
expected [ 2147483648 ] to include 2147483647
```

### DIFF
--- a/.changeset/settimeout-overflow.md
+++ b/.changeset/settimeout-overflow.md
@@ -1,0 +1,5 @@
+---
+'agenda': patch
+---
+
+Clamp near-future `setTimeout` delays to `2^31 - 1` ms. Node.js treats delays larger than a signed 32-bit integer as an overflow and clamps them to 1 ms, causing the JobProcessor to immediately reprocess a far-future job. The new ceiling matches Node's actual limit and prevents immediate reprocessing when `processEvery` is configured larger than that job's delay.
--- a/packages/agenda/src/JobProcessor.ts
+++ b/packages/agenda/src/JobProcessor.ts
@@ -14,7 +14,7 @@ const log = debug('agenda:jobProcessor');
 
 const agendaVersion = packageJson.version;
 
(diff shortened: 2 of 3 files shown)


santhiprakash:

PR #1800, opened 2026-07-24
fix(backends): don't double-count disabled jobs in getJobsOverview

### LINKED ISSUE #1768: getJobsOverview double-counts disabled (paused) jobs
### Bug

`getJobsOverview()` double-counts disabled (paused) jobs. For each disabled job, the per-state buckets sum to `total + (number of disabled jobs)` instead of `total`.

### Cause

`computeJobState()` (`packages/agenda/src/types/JobQuery.ts`) never returns `'paused'` — it derives state purely from timestamps. But each backend's `getJobsOverview()` loop does:

```js
const state = computeJobState(job, now);
overview[state]++;            // e.g. counts a disabled future job as "scheduled"
if (job.disabled === true) {
  overview.paused++;          // ...and also as "paused"
}
```

So a single disabled job is counted twice. This also makes the overview inconsistent with `queryJobs`, which treats `paused` (`disabled === true`) as a mutually exclusive state. Dashboards (agendash) that sum these buckets show incorrect totals.

Affected backends (identical pattern):
- `packages/mongo-backend/src/MongoJobRepository.ts`
- `packages/postgres-backend/src/PostgresJobRepository.ts`
- `packages/redis-backend/src/RedisJobRepository.ts`

### Expected

A disabled job should be counted only as `paused`. The per-state buckets should sum to exactly `total`.

### Fix

Count disabled jobs exclusively:

```js
if (job.disabled === true) {
  overview.paused++;
} else {
  const state = computeJobState(job, now);
  overview[state]++;
}
```

santhiprakash:
## Problem

`getJobsOverview()` double-counts disabled jobs. A disabled job with a future `nextRunAt` was counted once under `scheduled` (its computed state) and again under `paused` (because `job.disabled === true`).

Fixes #1768.

## Fix

In all three backend implementations (`mongo-backend`, `redis-backend`, `postgres-backend`), count a disabled job only under `paused` and do not increment its computed state.

- `MongoJobRepository.ts`
- `RedisJobRepository.ts`
- `PostgresJobRepository.ts`

## Verification

Added a shared test covering a disabled job with a future `nextRunAt` and verified it is reported as `paused: 1, scheduled: 0, queued: 0, completed: 0`.

- `pnpm --filter mongo-backend typecheck` passed
- `pnpm --filter mongo-backend test` passed (262 tests, 3 skipped)

### DIFF
--- a/packages/agenda/test/shared/repository-test-suite.ts
+++ b/packages/agenda/test/shared/repository-test-suite.ts
@@ -703,6 +703,29 @@ export function repositoryTestSuite(config: RepositoryTestConfig): void {
 				expect(overviewJob).toBeDefined();
 				expect(overviewJob!.total).toBe(2);
 			});
+
+			it('should not double-count disabled jobs as paused and their computed state', async () => {
+				// A disabled job with a future nextRunAt is reported only as paused,
+				// not as both paused and scheduled (see issue #1768).
+				const saved = await repo.saveJob({
+					name: 'paused-overview-job',
+					priority: 0,
+					nextRunAt: new Date(Date.now() + 60000),
+					disabled: true,
+					type: 'normal',
+					data: {}
+				}, undefined);
+				expect(saved.disabled).toBe(true);
+
+				const overview = await repo.getJobsOverview();
+				const overviewJob = overview.find(o => o.name === 'paused-overview-job');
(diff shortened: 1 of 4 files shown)


santhiprakash:

PR #1801, opened 2026-07-24
fix(agenda): make Job.unique() merge with existing uniqueOpts

### LINKED ISSUE #1762: Fluent unique() discards a previously-set debounce()/uniqueOpts
## Description

`Job.unique()` overwrites `uniqueOpts` wholesale:

```ts
unique(unique, opts?) {
  this.attrs.unique = unique;
  this.attrs.uniqueOpts = opts;   // clobbers existing opts
  return this;
}
```

`debounce()` merges its config into `this.attrs.uniqueOpts`. So calling `.debounce(d).unique(u)` silently drops the debounce config because `unique()` replaces `uniqueOpts` with `undefined`.

## Impact

`.debounce(2000).unique({ 'data.x': 1 })` saves a job with no debounce config — debouncing silently does not happen. Builder-method order should not matter here.

## Fix

Don't clobber existing opts when `unique()` is called without explicit options:

```ts
this.attrs.uniqueOpts = opts ?? this.attrs.uniqueOpts;
```

santhiprakash:
## Problem

`Job.unique()` replaces `this.attrs.uniqueOpts` with the new `opts` object. If the fluent chain called `.debounce()` first, the debounce options are silently discarded when `.unique()` runs.

Fixes #1762.

## Fix

Change `Job.unique()` to merge `opts` into the existing `this.attrs.uniqueOpts` instead of overwriting it.

## Verification

Added a unit test that calls `.debounce()` then `.unique()` and asserts that both `debounce` and `insertOnly` survive in `uniqueOpts`.

- `pnpm --filter agenda test` passed (187 tests).

### DIFF
--- a/packages/agenda/src/Job.ts
+++ b/packages/agenda/src/Job.ts
@@ -238,7 +238,7 @@ export class Job<DATA = unknown | void> {
 		opts?: JobParameters['uniqueOpts']
 	): this {
 		this.attrs.unique = unique;
-		this.attrs.uniqueOpts = opts;
+		this.attrs.uniqueOpts = { ...this.attrs.uniqueOpts, ...opts };
 		return this;
 	}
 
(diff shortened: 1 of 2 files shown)


santhiprakash:

